PR #1, opened 2018-11-26
fix: package improvements

wtrocki:
## Motivation

Cleanup ts-config
Move towards per package compilation
Add way to support running example app

## Outstanding work

- Watch
- Cleanup


.editorconfig
```diff
@@ -0,0 +1,9 @@
+root = true
```
darahayes: Minor question - is this really necessary? We don't really use this in other libs :)

wtrocki: We use that in all javascript project we have.  
This is to support peoples formatting practices across editors.
I see unformatted code everyday and with this single file you can get things standarized across the project.   Then I do not need to point every possible formating issue in reviews.

It is really saving a day connected with ts-lint for best practices.

wtrocki: Just example from 3 days ago: https://github.com/aerogear/aerogear-js-sdk/pull/160#discussion_r236002725

wtrocki: I can remove this fine to avoid flamewar (I will add that to gitignore and have it locally)

darahayes: Might as well leave it in there 😄

PR #4, opened 2018-11-28
feat: initial keycloak auth service

darahayes:
This PR is an initial port of the `KeycloakSecurityService` that was originally implemented in the [data-sync-server](https://github.com/aerogear/data-sync-server).

This initial implementation is fully working however there is outstanding work to be done. **Please just read these bullet points before jumping into the code changes**

* Integration & unit tests still need to be ported from the original implementation
* Need to improve the typings of certain things. For example, the `SecurityService` interface could have much stronger typing. @wtrocki I might get some help from you on this.
* Need to figure out how to configure / pass down a logger properly
* Need to review error classes being used inside the implementation
* Right now the KeycloakSecurityService is mostly set up directly in the end user code. Look at how the `schemaDirectives` and the `AuthContextProvider` are being used in `examples/keycloak/server.js` to get an idea. IMO this is not very tricky to set up from an end user perspective, but at a later stage we could simplify this and provide a better integration with our framework.

# Verification

An example has been provided under the `examples/keycloak` directory that shows how to get set up from scratch with a local keycloak server. The following instructions are copy-pasted directly from the new instructions I've added to the readme.

## Keycloak Example

`keycloak/server.js` shows how we can use the `KeycloakSecurityService` from `apollo-voyager-keycloak` to protect our app with Keycloak.

This example shows 

* How to set up authentication on the `/graphql` endpoint
* How to add role based access control on a Schema level using the `@hasRole` directive.

### Keycloak Setup

This example requires some extra setup. A `docker-compose.yml` file has been included to simplify a local keycloak setup.

The following steps set up a local keycloak instance, configures the instance for an example application and sets up a user that we can log into the application.

```
cd examples/keycloak/config
docker-compose up
```

* Open [http://localhost:8080/auth/admin/](http://localhost:8080/auth/admin/) and login with the user `admin` and password `admin`.
* Click **Add Realm** and click **Select File** next to the **Import** label.
* Select the [`examples/keycloak/config/realm-export.json`](examples/keycloak/config/realm-export.json) file and click **Create**.
* Click **Users** and add a new user called `developer`. You can choose your own name if you wish.
* Under the **Credentials** tab add a new password of **developer** and make sure it is not temporary. You can choose your own password if you wish.
* Under the **Role Mappings** tab assign the **admin** realm role.
* Select the **voyager-testing** option from the **Client Roles** dropdown and assign the **admin** role.

### Start the Server

```
node examples/keycloak/server.js
Initializing Keycloak authentication
🚀 Server ready at http://localhost:4000/graphql
```

Open [`http://localhost:4000/graphql`](http://localhost:4000/graphql) and you will be redirected to a login page. Log in with the user that was created earlier you should now see the the GraphQL playground.

In the playground you will see an error.

```json
{
  "error": "Failed to fetch schema. Please check your connection"
}
```

Do not worry, this error is caused by the playground making unauthenticated requests. One more step is needed.

In a new tab, open [`http://localhost:4000/token`](http://localhost:4000/token). You should see a JSON result.

```json
{"Authorization":"Bearer <Long String of Characters>"}
```

Copy the entire JSON result to your clipboard and navigate back to the Playground at [`http://localhost:4000/graphql`](http://localhost:4000/graphql). 

In the Playground, click the **HTTP Headers** button and paste the JSON result into the input box. If successful, the error will disappear and it is now possible to make queries.

Try out the following query

```
query hello {
  hello
}
```

### Role Based Access Control

The query above will only work if the authenticated user has the `admin` role. You can see this rule being applied with the `@hasRole(role: "admin")` directive in [examples/keycloak/server.js](examples/keycloak/server.js#L22).


darahayes:
@StephenCoady for some reason I cannot select you for review :/  Tt would be cool if you could try out the verification steps at some stage. No rush though.

psturc:
@darahayes Just tried this locally by following the instructions in readme and it worked perfectly 👍

darahayes:
@psturc awesome! Thanks for reviewing!

README.md
```diff
@@ -42,9 +42,93 @@ The `examples` directory has example scripts that show how the Voyager framework
 
 If you have run the `npm run bootstrap` command, the dependencies should already be installed.
 
-Try out `hello-server.js`.
+## Basic Example
+
+`basic/server.js` is the simplest example of how to use the `apollo-voyager-server` framework.
+
+```
+$ node examples/server.js
```
psturc: `$ node examples/basic/server.js`

wtrocki:
>  Need to improve the typings of certain things. For example, the SecurityService interface could have much stronger typing. @wtrocki I might get some help from you on this.

Happy to help here

> Need to figure out how to configure / pass down a logger properly 

See config interface in sync JS-SDK. It has all configuration passed to root level function

examples/keycloak/server.js
```diff
@@ -0,0 +1,84 @@
+const fs = require('fs')
+const path = require('path')
+const express = require('express')
+const { makeExecutableSchema } = require('graphql-tools')
+
+const { ApolloVoyagerServer, gql } = require('../../packages/apollo-voyager-server')
```
wtrocki: [Optional] We could use package name instead to support copy paste etc. 
Locally package will be linked so it should work fine

examples/keycloak/server.js
```diff
@@ -0,0 +1,84 @@
...
+const { KeycloakSecurityService } = require('../../packages/apollo-voyager-keycloak')
+
+
+const keycloakConfig = JSON.parse(fs.readFileSync(path.resolve(__dirname, './config/keycloak.json')))
+
+// This is our Schema Definition Language (SDL)
+const typeDefs = gql`
+
+  # In the older version of graphql in the data sync server
```
wtrocki: WOW. This means that users will need to define our directives?
Maybe we could concatenate user defined schema with our directives later.
Need to track this change upstream because I cannot believe that this is required. 
In previous version makeExecutableSchema accepted separate directives field.

packages/apollo-voyager-keycloak/src/AuthContextProvider.ts
```diff
@@ -0,0 +1,31 @@
+import { Request } from "express"
+import { IncomingMessage } from "http"
+
+export default class KeycloakAuthContextProvider {
```
wtrocki: I tend to avoid `default` as it will introduce difference between typescript and js usage.
For common js default import looks like `require("x").default which is missleading

packages/apollo-voyager-keycloak/src/AuthContextProvider.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  public readonly request : any
+  public readonly accessToken : any
+  public readonly authenticated: boolean
+
+  constructor (request: any) {
+    this.request = request
+    this.accessToken = (request && request.kauth && request.kauth.grant) ? request.kauth.grant.access_token : undefined
```
wtrocki: Looks like keycloak is not exposing this (seems to be internal api): https://github.com/keycloak/keycloak-js-bower/blob/master/dist/keycloak.d.ts

We should stick with any

packages/apollo-voyager-keycloak/src/KeycloakSecurityService.ts
```diff
@@ -0,0 +1,75 @@
...
+import KeycloakAuthContextProvider from './AuthContextProvider'
+import { Router } from 'express'
+import session from 'express-session'
+import Keycloak from 'keycloak-connect'
+
+export class KeycloakSecurityService implements SecurityService {
+  
+  public readonly keycloakConfig: any
+  public readonly schemaDirectives: any
+  public readonly authContextProvider: any
```
wtrocki: This can be mapped as interface.

packages/apollo-voyager-keycloak/src/schemaDirectives/hasRole.ts
```diff
@@ -0,0 +1,68 @@
+import { SchemaDirectiveVisitor } from 'graphql-tools'
+import { defaultFieldResolver } from 'graphql'
+import { ForbiddenError } from 'apollo-server-express'
+// import newInternalServerError from '???' // need to figure out where this comes from
+import pino from 'pino' // also need to figure out where this comes from
+import Joi from 'joi'
+
+const log = pino()
```
wtrocki: Do they recommend to setup instance for each file or we can have global logger?

StephenCoady:
@darahayes played with this and it all works as expected. really excellent! The only thing I noticed (playing the ignorance card here) is that after I remove the admin role from the developer I don't seem to be able to cause any errors. I know this is probably a Keycloak thing but is this because the token is still active regardless of when/to whom it was issued? How long after the user has logged in does it expire? Is the only way to revoke it to log out? I tried that and was still able to perform the query.

examples/keycloak/server.js
darahayes: `makeExecutableSchema` still accepts the separate directives field however, in `graphql` there is a function called `buildASTSchema` (which is used by `makeExecutableSchema`). This function does not take that directives field into account when validating the input typedefs. In the older version of `graphql`, there was way less validation.

It's possible that the old way was actually invalid and that this new behaviour is the intended behaviour. It might be worth opening an issue against `graphql-tools` because Apollo still have some example code in their docs using the old way.

darahayes: We could ask them if they know about this change. Is it the intended behaviour etc.

packages/apollo-voyager-keycloak/src/AuthContextProvider.ts
darahayes: Thanks I stopped doing it but there must be a few still lingering around :)

packages/apollo-voyager-keycloak/src/schemaDirectives/hasRole.ts
darahayes: I haven't seen any specific recommendation but because we're looking to provide things like context logging, we should probably provide some way to plug a logger in. Either for our framework to pass it down, or for the user to pass it down.

darahayes:
@StephenCoady I believe it's to do with the role info being stored inside the token so yes... when the token expires or when you log out it should stop working. Might be a silly question but did you remove the client role as well as the realm role?

In any case, we'll have to look into what the best practices are around this.

packages/apollo-voyager-keycloak/src/schemaDirectives/hasRole.ts
wtrocki: Frameworks tend to use debug package. Pino etc. are usually used in end user app

examples/keycloak/server.js
wtrocki: Just noticed that they examples aren't using latest version of graphql. Maybe they do that intentionally and there is recommended version set from apollo team. Anyway good to ask around

aliok:
> @darahayes played with this and it all works as expected. really excellent! The only thing I noticed (playing the ignorance card here) is that after I remove the admin role from the developer I don't seem to be able to cause any errors. I know this is probably a Keycloak thing but is this because the token is still active regardless of when/to whom it was issued? How long after the user has logged in does it expire? Is the only way to revoke it to log out? I tried that and was still able to perform the query.

This happens because Keycloak and our server is loosely coupled.

> (playing the ignorance card here)

Let me play along and provide long explanation :)

When you paste the token you received to https://jwt.io/, you can see something like this:
```
{
  "jti": "9be9b92a-611e-4399-be98-4121575a83f6",
  "exp": 1544736087,
  "nbf": 0,
  "iat": 1544700087,
  "iss": "http://localhost:8080/auth/realms/voyager-testing",
  "aud": "voyager-testing",
  "sub": "cfd5f04c-1e6a-4f6c-a48c-26f5ee874211",
  "typ": "Bearer",
  "azp": "voyager-testing",
  "auth_time": 1544700087,
  "session_state": "56dab35d-2808-4f9f-81e9-7ec788b89e1e",
  "acr": "1",
  "allowed-origins": [
    "*"
  ],
  "realm_access": {
    "roles": [
      "admin",
      "uma_authorization"
    ]
  },
  "resource_access": {
    "voyager-testing": {
      "roles": [
        "admin"
      ]
    },
    "account": {
      "roles": [
        "manage-account",
        "manage-account-links",
        "view-profile"
      ]
    }
  },
  "name": "Ali Ok",
  "preferred_username": "developer",
  "given_name": "Ali",
  "family_name": "Ok",
  "email": "aliok@aaa.com"
}
```
It has all the relevant information and all the roles/etc. (depending on your configuration)

A signature is also part of that token. If the signature is "valid", which means signed by Keycloak, we trust the token. The public key (or some secret) of the signing party is available on the Sync server and that's all you need for verifying a token's signature. Or, it is also the case that it is not available immediately but it is fetched from Keycloak (see realm-public-key section in https://www.keycloak.org/docs/latest/securing_apps/index.html#_java_adapter_config).

> I tried that and was still able to perform the query.

I think you didn't get a new token but logged out only and used the old token.
As the token verification happens in sync server side, it doesn't know that the token is actually now not up to date.

There are methods to take care of situations like this: https://github.com/lcobucci/jwt/issues/80

But, I don't think any of this is in sync server/framework scope.

BTW,

> Might be a silly question but did you remove the client role as well as the realm role?

You can even kill Keycloak and delete the whole realm etc. Result will be the same (I tried it)

Having said all of the things above, I am not an expert in any of this and there might be things I am wrong.
Let's ping @wei-lee and perhaps he can have a look after holidays.

aliok:
cc @wtrocki

aliok:
For future reference, this is the conclusion: https://issues.jboss.org/browse/AEROGEAR-8293

PR #5, opened 2018-11-29
feat: initial prometheus metrics port

darahayes:
This PR is pretty much a copy-paste of the old metrics module from the sync server.

One of the important metrics that needs to be collected is resolver timings. I imagine we will need to implement some sort of wrapper that the user can choose to enable as part of overall metrics. This PR does not include a solution for this.

My suggestion is that we merge this PR without verification as I have another PR around CircleCI that I want to get merged ASAP. Then we can work out the resolver timing solution.

### DIFF
--- a/packages/apollo-voyager-metrics/src/index.ts
+++ b/packages/apollo-voyager-metrics/src/index.ts
@@ -1,0 +1,96 @@
+import Prometheus from 'prom-client'
+import { buildPath } from '@aerogear/apollo-voyager-tools'
+import { IncomingMessage } from 'http'
+import { Response } from 'express'
+
+interface ResponseWithVoyagerMetrics extends Response {
+  requestStart: number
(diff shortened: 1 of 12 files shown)


packages/apollo-voyager-metrics/src/index.ts
```diff
@@ -0,0 +1,96 @@
...
+
+    const requestFailed = (err !== undefined || res.statusCode > 299)
+
+    serverResponseMetric
+      .labels(requestMethod, String(requestFailed))
+      .observe(responseTime)
+  }
+}
+
+export function updateResolverMetrics (resolverInfo: any, responseTime: number) {
```
wtrocki:
```suggestion
export const updateResolverMetrics = (resolverInfo: any, responseTime: number) => {
```

PR #6, opened 2018-11-30
ci: Initial CircleCI config

darahayes:
@wtrocki sorry for such a noisy PR. Baiscally, this gives us a working CI (without tests for now). I had to do some things to get this working.

* Individual packages do not have `package-lock.json` anymore.
* The top level folder does have a `package-lock.json`
* I've started using the `file:` specifiers for sibling dependencies.
* I've put `graphql` and `@types/graphql` as dev dependencies at the top level. This stops us from needing to do any hoisting.

Result:

* We have one top level package-lock.json that specifies all the deps
* Don't need to do any weird hoisting which causes failures in CI
* Individual packages do not have package-lock.json for production deps which is a good thing, however we do have a top level package-lock which ensures our dev dependencies are stable :)


packages/apollo-voyager-keycloak/package.json
```diff
@@ -23,7 +23,7 @@
     "test": "echo \"Error: run tests from root\" && exit 1"
   },
   "dependencies": {
-    "@aerogear/apollo-voyager-server": "1.0.0",
+    "@aerogear/apollo-voyager-server": "file:../apollo-voyager-server",
```
wtrocki: This will cause havoc on release time, however this could be swapped by lerna for release time. What is the context/problem for this change. Just for my information

tslint.json
```diff
@@ -1,10 +1,21 @@
...
   "extends": "tslint:recommended",
+  "linterOptions": {
+      "exclude": [
+          "node_modules/**"
+      ]
+  },
   "rules": {
+    "semicolon": [
+      true,
+      "never"
```
wtrocki: LOL

.circleci/config.yml
```diff
@@ -0,0 +1,73 @@
...
+
+workflows:
+  version: 2
+  build_and_release:
+    jobs:
+      - build:
+          filters:
+            tags:
+              only: /.*/
+      - npm_publish:
```
wtrocki: We can reuse scripts done js-sdk to support this case. Not sure if we want to copy this along with docs now or it will be separate ticket

.circleci/config.yml
```diff
@@ -0,0 +1,73 @@
...
+      - run:
+          name: install dependencies
+          command: npm install
+      - run:
+          name: bootstrap project
+          command: npm run bootstrap
+      - run:
+          name: compile
+          command: npm run compile
+      # - save_cache:
```
wtrocki: Does this has any value or are just experiments?

packages/apollo-voyager-keycloak/package.json
darahayes: Seeing it recommended in many places. It basically reduces the dependency on using the `lerna bootstrap` command.

https://github.com/lerna/lerna/issues/1449#issuecomment-414713331
https://github.com/lerna/lerna/issues/1462#issuecomment-397154543

I am going to test the old way again and see can I get it working.

.circleci/config.yml
darahayes: Separate ticket here: https://issues.jboss.org/browse/AEROGEAR-8181

PR #7, opened 2018-12-05
Initial test structure

aliok:
https://issues.jboss.org/browse/AEROGEAR-8183

Changes:
* Introduced ava in all packages and setup their test scripts and ava settings. For the packages that don't have any tests yet, I created a "deleteme.test.js" file so that ava doesn't complain
* Moved 2 tests: 
  * packages/apollo-voyager-tools/src/buildPath.test.ts
  * packages/apollo-voyager-keycloak/src/schemaDirectives/hasRole.test.ts
* Made lint command don't check tests
* But, made format command to also format tests (--> we could actually create a Git commit hook for this)

As per https://github.com/lerna/lerna#common-devdependencies

> Note that devDependencies providing "binary" executables that are used by npm scripts still need to be installed directly in each package where they're used.

we need to have ava and typescript as devDeps in subpackages.

CircleCI is happy.


aliok:
@darahayes @wtrocki let's merge this one and then I will do these separately:
- try setting up on-the-fly compilation for tests with Ava
- move some more tests

wtrocki:
@aliok This looks good, but there is chance that we missing git ignore for package lock files.
This will need to be rebased after @darahayes fix will be merged.
Apart from that change looks great. I would personally add test command into each package instead of `exit 0` but that is optional (it will not fail even when there are no tests so next PR's will focus on adding tests to those packages.

wtrocki:
Out of curiosity: Do we want to lint unit tests as well? I believe that this will cause so much hassle that it is not worth it. We can disable lint in separate PR (package.json filter), but wanted to trigger discussion

aliok:
@wtrocki 
> Apart from that change looks great. I would personally add test command into each package instead of exit 0 but that is optional (it will not fail even when there are no tests so next PR's will focus on adding tests to those packages.

Let me do that separately, if @darahayes and you are ok with the structure.

> Out of curiosity: Do we want to lint unit tests as well? I believe that this will cause so much hassle that it is not worth it. We can disable lint in separate PR (package.json filter), but wanted to trigger discussion

Honestly, part of me says they're code so we should lint them. But on the other hand I just want to have the ability to pass mock/inpartial objects as 'any' without complaints.
Let me disable linting for tests.

wtrocki:
I have done that in #9

aliok:
> Apart from that change looks great. I would personally add test command into each package instead of exit 0 but that is optional (it will not fail even when there are no tests so next PR's will focus on adding tests to those packages.

Nope, this is not the case:
```
✖ Couldn't find any files to test
```

I will add dummy files

aliok:
@darahayes @wtrocki this PR is ready for the final review.
https://github.com/aerogear/apollo-voyager-server/pull/9 is included here but with slight improvement.

package.json
```diff
@@ -2,11 +2,11 @@
...
   "scripts": {
-    "test": "echo ok",
-    "bootstrap": "lerna bootstrap --no-ci",
+    "test": "lerna exec -- npm test",
+    "bootstrap": "lerna bootstrap  --no-ci",
     "clean": "lerna clean --yes",
-    "lint": "tslint '*/*/src/**/*.ts' '*/*/test/**/*.ts'",
-    "format": "tslint '*/*/src/**/*.ts' '*/*/test/**/*.ts' --fix",
+    "lint": "tslint '*/*/src/**/*.ts' --exclude '*/*/src/**/*.test.ts'",
+    "format": "tslint '*/*/src/**/*.ts' --fix --force > /dev/null",
```
darahayes: Just wondering what's the `> /dev/null` thing for here? Is this going to suppress error output?

packages/apollo-voyager-context/src/deleteme.test.ts
```diff
@@ -0,0 +1,5 @@
+import test from 'ava'
+
+test('DELETE ME DUMMY', t => {
```
darahayes: 😄

packages/apollo-voyager-keycloak/src/schemaDirectives/hasRole.test.ts
```diff
@@ -0,0 +1,249 @@
+import test from 'ava'
+
+import {GraphQLSchema} from 'graphql'
```
darahayes: Really minor thing but just check spacing around imports like this. I'm wondering is this something your IDE does? I wonder can we lint for this.

package.json
aliok: Short answer yes:

Long answer
------------
2 scripts:
* `lint`: Complains with "ERROR: bla bla" messages, doesn't fix, exits with non-zero in case of error, only used in non-test code
* `format`: It runs on test and non-test code. Don't want to see "ERROR: bla bla" in the output because the tests might have too many of them, so that's why I put "> dev/null". I don't want it to return non-zero, so I put "--force".

packages/apollo-voyager-keycloak/src/schemaDirectives/hasRole.test.ts
aliok: Yeah, my IDE does format like that.
But I was assuming it is using eslint. Let me have a look at this ...tomorrow.

aliok: @darahayes I tried tslint's --fix now and it doesn't fix the spaces for that case.
I tried bunch of tslint rules to fix/enforce that behavior but I wasn't successful. 

UPDATE: I found this open issue on tslint: https://github.com/palantir/tslint/issues/1044

Anyway, I will fix this manually.

aliok:
> @aliok I'm personally -1 on trying to do formatting as part of a commit hook because the formatting will be run after you've already git added your changes and I don't think we should try to auto git add stuff because that would likely lead to adding the wrong things.

OK, I don't have a strong opinion in terms of Git hook.

> I recommend we add a separate lint config for tests. Eventually we will add a commit hook that only does linting but no formatting. What do you think? :)

I can try crteating a separate lint config, but in another PR. Let's merge this one first.

@darahayes give me your approval!

PR #8, opened 2018-12-05
fix: remove package-lock from packages

darahayes:
What it does:

* Removes `package-lock.json` from individual packages. <- Trust me [this is a good thing](https://github.com/sindresorhus/ama/issues/479#issuecomment-310661514).
* Adds the `--no-ci` flag to lerna bootstrap. This ensures `cipm` (the ci npm tool) is not used. `cipm` does not like when there is no `package-lock.json`


aliok:
Let's merge! Then I can rebase my PR on this one.

PR #9, opened 2018-12-05
fix: Disable lint for test that can lead to unnecessary improvements

wtrocki:
Linting source code is enough to guarantee quality. 
Linting tests can bring a lots of modifications when migrating tests from server.
We still need to keep format option to format source code, but let's not fail on errors in tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
     "test": "echo ok",
     "bootstrap": "lerna bootstrap",
     "clean": "lerna clean --yes",
-    "lint": "tslint '*/*/src/**/*.ts' '*/*/test/**/*.ts'",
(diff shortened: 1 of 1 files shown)


darahayes:
@wtrocki I'm not 100% sure if I see the value in this. The previous tests were written in JS which means someone will have to do the work of converting those to TypeScript. I would consider part of that work includes formatting the code correctly 😄 Wdyt?

wtrocki:
It's more about types: you need to supply a lots of any to mocks and other stuff. but yeah 
Feel free to close this

aliok:
Similar thing done in https://github.com/aerogear/apollo-voyager-server/pull/7 but with a bit more of improvements

PR #10, opened 2018-12-06
Aerogear 8183 test setup 2

aliok:
As discussed in https://github.com/aerogear/apollo-voyager-server/pull/7

Separate and more lenient tslint ruleset for tests.


aliok:
@darahayes  now that https://github.com/aerogear/apollo-voyager-server/pull/7 is merged, what do you think about this PR?

Shadowing is now allowed in tests. We can disable more rules if necessary when we're adding more tests in the future.
Part https://issues.jboss.org/browse/AEROGEAR-8183 is moving tests so I can figure out more stuff.

wtrocki:
This exactly what I meant! Had tons of problems with linting tests and this is much better than just disabling tests. Awesome change

PR #12, opened 2018-12-07
AEROGEAR-8186 Move existing prometheus metrics functionality from syn…

aliok:
…c server to its own library

As we decided, I picked the simplest approach.

Changes in examples/metrics/server.js is done to demonstrate how to make use of these changes.

Verificaiton:
1. Start it by `node examples/metrics/server.js `
2. ~~Execute this query bunch of times~~
```
query hello {
  getUser{name, favoriteHorse{name, breed}}
}
```
UPDATE: the metrics example is simplified now. Just execute `query hello {hello}`.
3. Go to http://localhost:4000/metrics
You will see these:
```
# HELP resolver_timing_ms Resolver response time in milliseconds
# TYPE resolver_timing_ms histogram
resolver_timing_ms_bucket{le="0.005",operation_type="query",name="Query.hello"} 0
resolver_timing_ms_bucket{le="0.01",operation_type="query",name="Query.hello"} 0
resolver_timing_ms_bucket{le="0.025",operation_type="query",name="Query.hello"} 0
resolver_timing_ms_bucket{le="0.05",operation_type="query",name="Query.hello"} 0
resolver_timing_ms_bucket{le="0.1",operation_type="query",name="Query.hello"} 0
resolver_timing_ms_bucket{le="0.25",operation_type="query",name="Query.hello"} 0
resolver_timing_ms_bucket{le="0.5",operation_type="query",name="Query.hello"} 0
resolver_timing_ms_bucket{le="1",operation_type="query",name="Query.hello"} 1
resolver_timing_ms_bucket{le="2.5",operation_type="query",name="Query.hello"} 1
resolver_timing_ms_bucket{le="5",operation_type="query",name="Query.hello"} 1
resolver_timing_ms_bucket{le="10",operation_type="query",name="Query.hello"} 1
resolver_timing_ms_bucket{le="+Inf",operation_type="query",name="Query.hello"} 1
resolver_timing_ms_sum{operation_type="query",name="Query.hello"} 1
resolver_timing_ms_count{operation_type="query",name="Query.hello"} 1

# HELP requests_resolved Number of requests resolved by server
# TYPE requests_resolved counter
requests_resolved{operation_type="query",path="Query.hello"} 1

# HELP requests_resolved_total Number of requests resolved by server in total
# TYPE requests_resolved_total counter
requests_resolved_total{operation_type="query",path="Query.hello"} 1

# HELP server_response_ms Server response time in milliseconds
# TYPE server_response_ms histogram
[...]

PR #13, opened 2018-12-10
Aerogear 8188 audit log module

aliok:
Blocked by https://github.com/aerogear/apollo-voyager-server/pull/12

Similar to https://github.com/aerogear/apollo-voyager-server/pull/12, a dumb audit logging implementation.

Created a new server startup in examples dir.

Verification:
- Run `npm run bootstrap && npm run compile`` on `apollo-voyager-server`
- Run `npm install` on `apollo-voyager-server/examples`
- Run `node apollo-voyager-server/examples/auditLogging/server.js`
- Go to localhost:4000/graphql and execute bunch of queries
- You will see some JSON output in the console.

### DIFF
--- a/examples/auditLogging/server.js
+++ b/examples/auditLogging/server.js
@@ -1,0 +1,48 @@
+const express = require('express')
+const { makeExecutableSchema } = require('graphql-tools')
+
+const { ApolloVoyagerServer, gql } = require('../../packages/apollo-voyager-server')
+const { setAuditLogEnabled, wrapResolversForAuditLogging} = require('../../packages/apollo-voyager-audit')
+
+// This is our Schema Definition Language (SDL)
+const typeDefs = gql`
+  type Query {
+    hello: String
+  }
+`
+
+// Resolver functions. This is our business logic
+let resolvers = {
+  Query: {
+    hello: (obj, args, context, info) => {
+      return `Hello world from ${context.serverName}`
+    }
+  }
+}
+
+setAuditLogEnabled(true)
+resolvers = wrapResolversForAuditLogging(resolvers)
+
+const schema = makeExecutableSchema({ typeDefs, resolvers })
+
+// The context is a function or object that can add some extra data
+// That will be available via the `context` argument the resolver functions
+const context = async ({ req }) => {
+  // add some context to GraphQL requests
+  console.log('my context provider')
+  return { serverName: 'Voyager Server' }
+}
+
+// Initialize the apollo voyager server with our schema and context
+const server = ApolloVoyagerServer({
+  schema,
+  context
+})
+
+const app = express()
+server.applyMiddleware({ app })
+
+const port = 4000
+app.listen({ port }, () =>
+  console.log(`🚀 Server ready at http://localhost:${port}${server.graphqlPath}`)
+)
--- a/packages/apollo-voyager-audit/src/index.ts
+++ b/packages/apollo-voyager-audit/src/index.ts
@@ -1,0 +1,75 @@
+import { buildPath } from '@aerogear/apollo-voyager-tools'
+import { GraphQLResolveInfo } from 'graphql'
+import { IFieldResolver } from 'graphql-tools'
+import pino from 'pino'
+
+const log = pino()
+const auditLogger = log.child({tag: 'AUDIT'})
+
+let auditLogEnabled = false
+
+export function isAuditLogEnabled (): boolean {
+  return auditLogEnabled
+}
+
+export function setAuditLogEnabled (val: boolean): void {
+  auditLogEnabled = val
+}
+
+interface ResolverObject {
+  [key: string]: IFieldResolver<any, any>
+}
+
+export function wrapResolversForAuditLogging (resolverMappings: { [key: string]: ResolverObject }): { [key: string]: ResolverObject } {
+  const output: { [key: string]: ResolverObject } = {}
+
+  const typeKeys = Object.keys(resolverMappings)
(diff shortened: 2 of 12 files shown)


examples/auditLogging/server.js
```diff
@@ -0,0 +1,48 @@
...
+// Resolver functions. This is our business logic
+let resolvers = {
+  Query: {
+    hello: (obj, args, context, info) => {
+      return `Hello world from ${context.serverName}`
+    }
+  }
+}
+
+setAuditLogEnabled(true)
```
wtrocki: Line 23/24: I guess we are working now to get that into core and pass config?

packages/apollo-voyager-audit/src/index.ts
```diff
@@ -0,0 +1,75 @@
...
+
+interface ResolverObject {
+  [key: string]: IFieldResolver<any, any>
+}
+
+export function wrapResolversForAuditLogging (resolverMappings: { [key: string]: ResolverObject }): { [key: string]: ResolverObject } {
+  const output: { [key: string]: ResolverObject } = {}
+
+  const typeKeys = Object.keys(resolverMappings)
+  for (const typeKey of typeKeys) {
```
darahayes: Seeing as we have this exact thing in two places, it might make sense to move it into `apollo-voyager-tools` at some stage. It could accept the wrapper function as an argument and look after the rest

examples/auditLogging/server.js
aliok: actually I don't...
as we spoke during the chat today, I went with the simplest and quickest way to progress on cards.

packages/apollo-voyager-audit/src/index.ts
```diff
@@ -0,0 +1,75 @@
...
+      output[typeKey][fieldKey] = wrapSingleResolverForAuditLogging(resolverForField)
+    }
+  }
+
+  return output
+}
+
+export function auditLog (success: boolean, request: any, info: GraphQLResolveInfo, parent: any, args: any, msg: string) {
+  if (auditLogEnabled) {
+    auditLogger.info({
```
wtrocki: [Optional] Define this as interface so we have automatically document in the code what is being logged and when.

packages/apollo-voyager-audit/src/index.ts
```diff
@@ -0,0 +1,75 @@
+import { buildPath } from '@aerogear/apollo-voyager-tools'
+import { GraphQLResolveInfo } from 'graphql'
+import { IFieldResolver } from 'graphql-tools'
+import pino from 'pino'
+
+const log = pino()
```
wtrocki: [Question] Any reason we use pino instead of plain console.log? Message formatting?

examples/auditLogging/server.js
wtrocki: That is no problem. We could do that in no time after this is merged.

packages/apollo-voyager-audit/src/index.ts
aliok: We need to log in JSON format for audit logging and we liked pino back then because it logs in JSON by default and it is efficient.
I simply moved over stuff from the old code base.

wtrocki:
This looks good. From my point of view I will add minimal information into individual packages README files.
[Optional] Unit tests for package (just simple ones to not have package without tests.

darahayes:
@aliok @wtrocki I am happy enough with this PR for now. We are still not solving the resolver composition problem but I feel like it's still a good idea to solve that separately.

I've had an idea for doing that but I'm not sure how you will feel about it. I think so far we have been talking about some kind of infinitely flexible resolver composition mechanism that the user can manipulate at a framework level. Something similar to the idea of express middlewares. But with the added complexity that we want to slot in things like resolver timings and audit logging at certain positions of that resolver chain.

I'm starting to feel like this is very complex and that it's a straight up bad idea. Instead I'm thinking we actually wrap all user resolvers with a single framework resolver that looks after our framework level stuff in the exact order we want, and it simply calls the user resolver somewhere in there. It's a rough idea but the steps would be something like this.

```
function voyagerResolver(obj, args, context, info) {
	// start resolver timing thing
	// do any pre resolver audit logging if necessary
	// call user level resolver
    // audit log of resolver details, failure, success etc.
    // collect and update resolver timing metrics
    // return
}
```

I see some massive benefits in this

* It's ridiculously simple. Do we actually need to provide a composition mechanism?
* The user can still combine their own resolvers at an application level using external tools such as the [`combine-resolvers`](https://github.com/lucasconstantino/graphql-resolvers) we mentioned. As far as we are concerned the user resolver will still be just one function.
* It's probably more performant and any resolver performance issues would be likely caused by poor user behaviour (combining loads of stuff at an app level) as opposed to our framework enabling it.
* Again - will the users need more than this? If they do, perhaps we can give them some limited access to modify how this wrapper function executes in the form of some lifecycle hooks type mechanism. for example pre and post user resolver execution hooks.

I feel like this may be a good approach to go. Apollo and `graphql` are not providing a composition mechanism so why should we? I would love to get your thoughts.

wtrocki:
@darahayes Fully agree with this. We could get this over the line (merge to master as it is) and refactor current code once we will get this resolver done as part of the server.

aliok:
@darahayes I am +1 on your idea. For me the key part is this:

> The user can still combine their own resolvers at an application level using external tools

Let's do it separately.

aliok:
@darahayes @wtrocki addressed the comments with a new commit. Please have a look and we can merge this one too.

wtrocki:
Verifying now.

PR #14, opened 2018-12-11
AEROGEAR-8246 - voyagerResolver

aliok:
https://issues.jboss.org/browse/AEROGEAR-8246

Changes:
- No more separate resolver wrapping for auditing and metrics

### DIFF
--- a/examples/auditLogging/server.js
+++ b/examples/auditLogging/server.js
@@ -1,8 +1,7 @@
 const express = require('express')
 const { makeExecutableSchema } = require('graphql-tools')
 
-const { ApolloVoyagerServer, gql } = require('../../packages/apollo-voyager-server')
-const { setAuditLogEnabled, wrapResolversForAuditLogging} = require('../../packages/apollo-voyager-audit')
+const { ApolloVoyagerServer, gql, voyagerResolvers } = require('../../packages/apollo-voyager-server')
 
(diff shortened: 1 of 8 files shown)


darahayes:
@aliok I am actually pretty happy with this implementation for now. I see some things changing eventually. Things like the `auditLog` implementation and the metrics implementation would potentially be passed down I imagine. But we agreed that we should solve that separately.

I've started on a very rough implementation of what the bigger picture might eventually look like here: https://github.com/aerogear/apollo-voyager-server/pull/16.

@aliok my only one small ask is could you move the resolver implementation and all of its imports into a separate file? I see these pieces (both server and resolvers) needing refactoring very soon and it would be nice if they were separated :)

wtrocki:
Really good stuff. Reviewed code.  @darahayes was faster with verification :)

aliok:
@darahayes 

> Things like the auditLog implementation and the metrics implementation would potentially be passed down by the user in a config object I imagine. But we agreed that we should solve that separately.
I've started on a very rough implementation of what the bigger picture might eventually look like here: #16.

Yes. Makes sense. I will also try to contribute there.

> @aliok my only one small ask is could you move the resolver implementation and all of its imports into a separate file? I see these pieces (both server and resolvers) needing refactoring very soon and it would be nice if they were separated. Happy to approve after that :)

Done

PR #16, opened 2018-12-11
breaking: refactor context into server module

darahayes:
### Description

The bigger picture is that the user will pass the specific implementation for various features as a config that goes into `ApolloVoyagerServer`. For example, the user will create a `KeycloakSecurityService` and that will be passed into `ApolloVoyagerServer` which will know how to set everything up internally. We should be doing this for audit logging, metrics, and any feature that we implement.

This specific PR does the following:

* Moves the behaviour of the `apollo-voyager-context` module into the `apollo-voyager-server` module.
* Adds a working example of passing the `KeycloakSecurityService` into the server using the `voyagerConfig`. Internally, the server can add the appropriate `KeycloakAuthContextProvider` into the GraphQL context.

Next Steps:

* Right now the `getContext` function is checking for an `AuthContextProvider` at runtime.
* If I understood @wtrocki properly, instead we should come up with some default configs and empty implementations that can be merged with whatever the user passes in. This would mean the `getContext` function does not have to perform these runtime checks and the behaviour in there is a lot simpler. It also provides us with opportunities to provide better typings on the `context` object.

@wtrocki can you confirm if my understanding is correct? If so, that is what I will focus on next!

##### Checklist

- [ ] `npm test` passes
- [ ] tests are included


wtrocki:
@StephenCoady Worth checking this PR

darahayes:
^^ don't look at it as an example of good code though. It's quite rough 😄

wtrocki:
The idea is to release burden from users to know what elements are passed to context and benefit from pluggable elements in the context. This means that no matter what config values will be passed context methods should not change (with some exceptions for auth etc we can add in the future)..

## How this could work

- [x]  server constructor accept config 
- [  ] server checks if user passed implementation into config (if not some default noop or other implementation should be mounted into context.
- User context should be also mounted into separate namespace to avoid clashes. 
- We will need to investigate if we can provide type for context object so users will get instantly all the functions that are available for it (this will be cool but not sure if we actually can do it as we will probably need to redefined apollo resolver function signature.

examples/keycloak/server.js
```diff
@@ -56,17 +55,22 @@ const schema = makeExecutableSchema({
...
-    // add the keycloak Auth Context Provider
-    auth: new AuthContextProvider(req) 
+    serverName: 'Voyager Server'
   }
 }
 
 // Initialize the apollo voyager server with our schema and context
-const server = ApolloVoyagerServer({ 
+
+const apolloConfig = {
```
wtrocki: How about we will have this inside our config. Just idea - not strong opinion

darahayes:
@wtrocki Thanks! we're definitely on the same page so 🎉

packages/apollo-voyager-keycloak/src/AuthContextProvider.ts
```diff
@@ -7,9 +7,9 @@ export class KeycloakAuthContextProvider {
   public readonly accessToken: any
   public readonly authenticated: boolean
 
-  constructor (request: any) {
-    this.request = request
-    this.accessToken = (request && request.kauth && request.kauth.grant) ? request.kauth.grant.access_token : undefined
+  constructor ({ req }: { req: any }) {
```
wtrocki: @types/express can give us request types.

packages/apollo-voyager-server/src/context/VoyagerContextConfig.ts
```diff
@@ -0,0 +1,6 @@
+import { SecurityService } from '../security/SecurityService'
+
+export interface VoyagerContextProviderConfig {
```
wtrocki: [Question] Is the plan for this to be independent from actual config or part of the config that user passes?

IMHO we need only 2 types:

- Config Interface with all flags and strategies for functionalities
- Context interface we going to strap to resolver function.

This one is extra and it seems to be used internally

darahayes: Yeah It does look like some duplication right now but I guess the idea is that the top level config is probably going to grow to something pretty sizeable and IMO it's probably not good to pass around these massive objects filled with irrelevant properties into the smaller components.

I suppose we could just pass the individual bits as arguments instead of passing objects 😃

wtrocki: Ah.. I was hoping we going to namespace that towards features like security/offline.
For example:
https://github.com/feedhenry-raincatcher/raincatcher-core/blob/master/demo/server/config-dev.js#L12-L32

darahayes: @wtrocki possibly we will do this. Hoewever, in lots of cases the user is passing some config into the implementation first. e.g. new `KeycloakSecurityService(keycloakServiceConfig)` so that item is now completely configured. Then when they pass the security service to `ApolloVoyagerServer` they hopefully shouldn't need to pass any extra things. I don't see namespacing as being too important right now because it would just end up something like this:

```js
{
  security: {
    service: myInitializedKeycloakService  
  }
}
```

^^ and there's no other stuff in there because that service should have all of the appropriate config inside it.

Did I misunderstand you or maybe miss something? or does that make sense?

wtrocki: I was thinking that services/packages could be stateless/helpers or deliver some functionalities and config will contain some specific entry points for those. WIth this users will work with only one config passed to server that can have all moving points. Then inside server we will mount different packages and stuff. 

This aproach is also good. Not sure if there is much difference apart from getting more granular control and separation: server making some individual choices and packages provide unopiniated functionality. This way we can achieve layers of abstraction and give users more powerful (kinda separation of responsibilties)

## TL;DR 

it's 
```
{
  security: {
    service: myInitializedKeycloakService  
  }
}
```
vs
```
{
  security: {
    directiveProvider?: myCustomOne  
  }
}
```

``` // In server config processor 
if(!security.directiveProvider){
  security.directiveProvider = new KeycloakDirectiveProvider();
}

```

I have no strong opinion about this, but I was initially thinking that we going to do it this way.

wtrocki: Actually changing my mind. If individual packages will provide types like 

```
security: {
    unauthorizedAccess: true
    directiveProvider?: myCustomOne  
  }
``` 

Then they can be mounted into top level config and provide the same value. 
The only difference is that in my proposition noop implementations will be supplied inside server, while in proposition above users need to pick the right class instantiate it. 

I really do not like to make this an issue and we could get this done one way and evaluate that in our sample server to see how it feels for user etc.

packages/apollo-voyager-keycloak/src/AuthContextProvider.ts
```diff
@@ -1,31 +1,27 @@
...
-    return this.isAuthenticated() ? this.getToken().content : null
+  public hasRole(role: string): boolean {
+    return this.isAuthenticated() && this.accessToken.hasRole(role)
   }
 
-  public hasRole (role: string) {
-    return this.isAuthenticated() && this.getToken().hasRole(role)
+  // TODO where can we get this from?? The accessToken is kind of like a user but not exactly.
+  public getUser() {
+    return null
```
darahayes: @wtrocki any ideas about this one?

packages/apollo-voyager-keycloak/src/schemaDirectives/hasRole.ts
```diff
@@ -26,25 +25,25 @@ export class HasRoleDirective extends SchemaDirectiveVisitor {
     const { resolve = defaultFieldResolver } = field
     const { error, value } = this.validateArgs()
 
+    if (error) {
+      log.error(`Invalid hasRole directive on field ${field.name}`, error)
```
darahayes: Just a note on this error check. It's checking that the `hasRole` args are valid. Previously it was inside the `field.resolve` function. This was to stop the `data-sync-server` crashing at startup time when someone misconfigured a `hasRole`. Instead, an error would be sent back to a client at runtime if they tried to call a query/mutation that had an invalid `hasRole`.

Now that we're in this framework mode, the previous error handling logic is a bad idea. This change ensures an error is thrown as soon as the directive is loaded if there is an issue.

packages/apollo-voyager-keycloak/src/AuthContextProvider.ts
wtrocki: Checked and it's not possible to do it. Users will be able to get email from kauth and then perform queries, but rest of the data is missing. There is separate endpoint to get all principal data, but technically application itself can have access to their own user storage.

packages/apollo-voyager-server/src/config/VoyagerConfig.ts
```diff
@@ -1,9 +1,12 @@
...
 
-export interface VoyagerConfig {
+export interface VoyagerConfig extends Config {
 
   /**
    * Optional security service to be applied
    */
   securityService: SecurityService
+
+  auditLog: any
```
wtrocki: ??

packages/apollo-voyager-server/src/security/AuthContextProvider.ts
```diff
@@ -0,0 +1,7 @@
+
+
+export interface AuthContextProvider {
+  isAuthenticated(): boolean
+  hasRole(role: string): boolean
+  getUser(): any // User???
```
wtrocki: Technically we could do `getUserId`

wtrocki:
Pretty much happy with the changes.

packages/apollo-voyager-keycloak/src/AuthContextProvider.ts
darahayes: @wtrocki This is kind of what I was thinking. Should I bring back `getToken()` for the time being?

darahayes:
Update: I've added some extensive unit tests around the DefaultApolloVoyagerConfig and around the ApolloAuthContextProvider. I hope these cover all of the cases :) cc @wtrocki @aliok

packages/apollo-voyager-server/package.json
```diff
@@ -20,7 +20,7 @@
     "compile": "tsc --build tsconfig.json",
     "watch": "tsc --build tsconfig.json --watch",
     "compile:clean": "tsc --build tsconfig.json --clean",
-    "test": "ava '*.test.ts' '**/*.test.ts'"
+    "test": "ava -v '*.test.ts' '**/*.test.ts'"
```
aliok: Good idea having `-v` in package.json scripts.
But, I think we should do the same in other packages too, to be consistent.

aliok:
@darahayes I am happy with these changes. Let's target merging this as soon as possible as it might block other work. Shall I start with verifications or do you have some other stuff to change?

darahayes:
@aliok You can start verification, just going to address minor feedback items. Basically, you just need to see if the examples still run. I'm getting the feeling that these examples could be turned into some kind of automated smoke tests.

darahayes:
@wtrocki I got rid of the `getUser` function from the `AuthContextProvider` interface for now. We will need to revisit that later or perhaps for the time being, we will need to accept a slightly leaky abstraction and allow a `getToken` kind of function. What do you think?

wtrocki:
@darahayes Checked that again. I believe that what we good is really good. 
I would prefer to get that merged as we have here couple good examples how to build features for server 
that we may need.

PR #18, opened 2018-12-13
Aerogear 8192 keycloak module migration

aliok:
Make Keycloak related things conform to a framework approach as opposite to a product approach where things happen magically.

Changes:
- As we have a framework now, security service doesn't return all the things necessary. Instead user can compose stuff.
- Tests for KeycloakAuthContextProvider

UPDATE: New ticket for this PR: https://issues.jboss.org/browse/AEROGEAR-8272


darahayes:
Hey @aliok I'm not 100% sure if this is the way we wanted to go. I think that `getSchemaDirectives` and `getAuthContextProvider` are important to have within the SecurityService interface. The idea was that the user could simply initialize the `KeycloakSecurityService` and pass it down into `ApolloVoyagerServer` and then ApolloVoyagerServer will be able to perform some setup bits using `getSchemaDirectives` and `getAuthContextProvider`. I'm really sorry I didn't communicate this well enough to you.

I've got this WIP PR here and sort of shows the direction we were thinking of going in https://github.com/aerogear/apollo-voyager-server/pull/16

It's a pretty big PR but if you take a look at the keycloak example server in there you should get the idea very quickly. The next steps in there are provide a layer on top of makeExecutableSchema, probably inside of `ApolloVoyagerServer`. This will remove the need for the user to directly call `getSchemaDirectives` and do things like manually wrapping resolvers. Please let me know if I didn't explain it well!

aliok:
@darahayes 

> user could simply initialize the KeycloakSecurityService and pass it down into ApolloVoyagerServer and then ApolloVoyagerServer will be able to perform some setup bits using getSchemaDirectives and getAuthContextProvider

This is the current code in examples/keycloak though:
```
const { KeycloakSecurityService } = require('../../packages/apollo-voyager-keycloak')
...
const keycloakService = new KeycloakSecurityService(keycloakConfig)
const AuthContextProvider = keycloakService.getAuthContextProvider()
const schemaDirectives = keycloakService.getSchemaDirectives()
...
const schema = makeExecutableSchema({ 
  ...
  schemaDirectives
})
...
const context = ({ req }) => {
  return { 
    ...
    auth: new AuthContextProvider(req) 
  }
}
```

So, the user still has to get stuff and pass them to some other places.

I haven't checked your code at #16 yet but above was the reason that I have done it that way.

You don't have to reply this comment. We can discuss again once I check your PR.

aliok:
Ok, @darahayes changed quite a lot in his PR #16  and I liked the direction he's going.
What I did in this PR was trying to solve some of the problems he also saw but in a different way :)

I will keep this PR open for a while and once we finishes his PR, I will revert the refactoring I made here and only rebase the tests I migrated.

aliok:
Won't merge

PR #20, opened 2018-12-13
Add some offline/conflicts resolution tests

danielpassos:


danielpassos:
@StephenCoady I was playing around your PR and added some tests to play with. Feel free to use or reject. It's in the upstream branch if you need to change something.

PR #38, opened 2018-12-28
How to import Kibana dashboard documentation

aliok:
https://issues.jboss.org/browse/AEROGEAR-8321


doc/guides/metricsAndAuditLogs.md
```diff
@@ -0,0 +1,69 @@
...
+Error: Importing AeroGear Data Sync - top level execution per platform - aaa (top_level_execution_per_platform_49f9a0b6-09b5-11e9-9597-069f7827c758) failed: Could not locate that index-pattern-field (id: audit.clientInfo.data.device.platform.raw)
+Error: Could not locate that index-pattern-field (id: audit.clientInfo.data.device.platform.raw)
+``` 
+
+Because of these conditions, Kibana saved objects have to be imported after there is some audit logs are already in ElasticSearch.
+At the moment, no mechanisms provided to overcome this problem.
+
+## Viewing the dashboard and audit logs
+
+When the Kibana saved objects are imported, a dashboard will be available with lots of different visualizations.
```
darahayes: Going to point out two minor things here and they might seem very nitpicky and specific but I'm only pointing these out because I think you can learn from it in the future. These are both things that native English speakers do all the time, so I hope you don't take it the wrong way!

Usually phrases like `'lots of <x>'` or `'lots of different <x>'` would be inappropriate in formal documents. It also doesn't describe what the visualisations are / what their purpose is.

You could replace this with something like 

`A dashboard is available with <several | a number of> visualizations <that show x | that display x | relating to x>.`

Also another really specific thing which even native English speakers mess up all the time is mixing tenses together in the same sentence. In the sentence above:

* `When the Kibana saved objects *are* imported` <- present tense, they are imported right now
* `A dashboard *will* be available with ...` <- future tense, something will happen.

The second part should probably be `A dashboard is available.`

You see this kind of thing all the time in technical docs. 

* `When you have done x, y will be available.` 
* `When x is true, y will be true also.`

It's not a hard and fast rule, but in most cases the second part of the sentence should match the tense of the first part of the sentence e.g. `When x is true, y is true also`. This usually feels more natural for the reader.

Sorry for the grammar lesson 😄

doc/guides/metricsAndAuditLogs.md
```diff
@@ -0,0 +1,69 @@
+# Displaying metrics and exploring audit logs
```
darahayes: Minor thing but headings should always be captialized with the exception of words like `the, and, of, for, in, at` etc. (unless they are at the beginning of the heading of course).

It should be `Displaying Metrics and Exploring Audit Logs`

doc/guides/metricsAndAuditLogs.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Prerequisites
+
+* Audit logging is enabled in Voyager server and also in the clients
+* OpenShift logging is enabled
+
+## Importing Kibana saved objects
```
darahayes: see above comment about capitalization

doc/guides/metricsAndAuditLogs.md
```diff
@@ -0,0 +1,69 @@
...
+
+```
+Error: Importing AeroGear Data Sync - top level execution per platform - aaa (top_level_execution_per_platform_49f9a0b6-09b5-11e9-9597-069f7827c758) failed: Could not locate that index-pattern-field (id: audit.clientInfo.data.device.platform.raw)
+Error: Could not locate that index-pattern-field (id: audit.clientInfo.data.device.platform.raw)
+``` 
+
+Because of these conditions, Kibana saved objects have to be imported after there is some audit logs are already in ElasticSearch.
+At the moment, no mechanisms provided to overcome this problem.
+
+## Viewing the dashboard and audit logs
```
darahayes: See above comment about capitalization

doc/guides/metricsAndAuditLogs.md
aliok: Thanks @darahayes 
This kind of stuff is a bit hard for me honestly :) I would rather leave it to native speakers, or even better --> docs team :)
Anyway, read the entire document and tried to fix whatever I found.

doc/guides/metricsAndAuditLogs.md
aliok: +1 done

PR #39, opened 2019-01-02
breaking: refactor config

darahayes:
### Description

The main purpose of this PR is to allow the user to pass in implementations for auditLogging and metrics into `ApolloVoyagerServer` via the `VoyagerConfig` object.

Take a look at the changes made to the example servers to see how this works.

* If we go down this route the easiest way to do it will be to force the user to pass  as part of the Apollo Config. This is a pretty big change because it means that the user no longer calls `makeExecutableSchema`. I'd love to get other peoples' opinions on this.

Some notable changes.

* user now passes `typedefs` and `resolvers` as part of `ApolloConfig`. no more makeExecutableSchema.
* user passes `metrics` and `auditLogger` as part of VoyagerConfig. This enables those features and allows users to pass custom implementations.
* user no longer has to get security `schemaDirectives` and pass them in. This is now handled under the hood. See the keycloak example to understand this.
* user has access to `context.auditLog`. It is still available (no op) even when an auditLogger implementation is not passed.
* audit logging functions have been refactored a little bit. The order of arguments have been switched up a bit.

## Verification

* Run each of the examples and ensure they work as expected
* Read through the code of the auditLogging and Metrics examples to understand how things have changed and been simplified.
* @aliok can you pay special attention to the audit logs and verify that they look good from your point of view?

##### Checklist

- [ ] `npm test` passes
- [ ] tests are included


examples/auditLogging/server.js
```diff
@@ -1,7 +1,8 @@
 const express = require('express')
 const { makeExecutableSchema } = require('graphql-tools')
 
-const { ApolloVoyagerServer, gql, voyagerResolvers } = require('../../packages/apollo-voyager-server')
+const { ApolloVoyagerServer, gql } = require('../../packages/apollo-voyager-server')
```
wtrocki: Sums up refactoring support in VSCode ;)

examples/metrics/server.js
```diff
@@ -1,8 +1,10 @@
 const express = require('express')
 const { makeExecutableSchema } = require('graphql-tools')
```
wtrocki: should be gone as well

wtrocki:
> if we go down this route the easiest way to do it will be to force the user to pass typedefs

I do not want to be opinionated but `let's do this!`.
Thing is to actually allow to pass all parameters that go to this methods (currently we missing directives array)

> I will probably make sense to put the auditLog function in there eventually.

This is something I'm dying to add to example app so if @aliok do not have any concerns we could do this :)

darahayes:
@wtrocki you are totally right. Going to get some documentation as part of this PR and we'll be in a really good place :)

packages/apollo-voyager-server/src/audit/DefaultAuditLogger.ts
```diff
@@ -0,0 +1,13 @@
+import { GraphQLResolveInfo } from 'graphql'
+import { AuditLogger } from './AuditLogger'
+
+export class DefaultAuditLogger implements AuditLogger {
+
+  public logResolverCompletion (msg: string, success: boolean, obj: any, args: any, context: any, info: GraphQLResolveInfo): void {
```
wtrocki: It actually begs to have interface and have ? on the parameter that is different. Then we can reuse. No strong opinion on it, but it will be much cleaner.

darahayes: As I mentioned before, there is a very important separation between these two functions. The user should not be able to log a resolver completions. They should not be able to pass the `success` thing at all.

I see these as two functions that serve two totally different purposes.

darahayes:
Merged this PR without the appropriate docs changes but I'm working on those immediately.

packages/apollo-voyager-server/src/audit/DefaultAuditLogger.ts
aliok: @darahayes what's the effect on Kibana dashboards now?

Confirm what I understood: "success" is the field we should look for existence for the resolver execution graphs such as "number of queries executed".

aliok: not the best place to ask this question :)

PR #42, opened 2019-01-04
CI tests of basic example apps

psturc:
### Description

https://issues.jboss.org/browse/AEROGEAR-8330
CI tests for example apps in this repo

### Notes
* moved example apps' server initialisation to a [separate files](https://github.com/aerogear/apollo-voyager-server/pull/42/files#diff-26e773147cd88f895ef325f99f3a3851) (thanks @darahayes for suggesting this!) and reflected this in [readme](https://github.com/aerogear/apollo-voyager-server/pull/42/files#diff-700cc2ac4031647f10e14c03cbd582d2)
* keycloak instance configuration part is mostly copied from [here](https://github.com/aerogear/data-sync-server/blob/master/integration_test/auth.integration.test.keycloak.js)
* changed the command to run the test [here](https://github.com/aerogear/apollo-voyager-server/pull/42/files#diff-6621de04c1c88756164d80c93cc2369dR10), so these tests are not run automatically with **lerna** in CI during build, because they require additional configuration (keycloak) - that's why I decided to put them into separate job

##### Checklist

- [x] test of basic example app
- [x] test of keycloak example app
- [x] all changes reflected in docs


darahayes:
@psturc so sorry I accidentally closed the PR.

This PR is fantastic, I just have one minor ask. Could you replicate the what you did with `basic.test.js` for the metrics, auditLog and conflicts example also? You might initially think that this does not have much value but it's more testing to ensure that each example is runnable, does not crash, and that the query provided does actually work. Kind of like a smoke test :)

wtrocki:
Awesome changes. Really like the fact that we have all cases covered.
-1 for `examples/package-lock.json` file for examples. We got rid of this from the packages anyway.

PR #45, opened 2019-01-07
WIP: State persistence

wtrocki:
## Motivation:
To resolve conflicts users sometimes will need to know the state before both modifications were happening. This will enable developers to see differences in their api. StatePersistence is an really simple abstraction that users can reuse to provide conflict base. Interface is optional giving users additional capabilities without forcing them to implement Persistence layer. Example persistence layer could be in memory or utilize database.

## Progress 

- [x] Implementation
- [ ] Tests
- [ ] Documentation
- [ ] Example app updates

### DIFF
--- a/packages/apollo-voyager-conflicts/src/api/ObjectConflictError.ts
+++ b/packages/apollo-voyager-conflicts/src/api/ObjectConflictError.ts
@@ -33,10 +33,10 @@ export interface ConflictData {
  * Error specific to Voyager framework
  */
 export class ObjectConflictError extends GraphQLError {
-  public conflictInfo: any
+  public conflictInfo: ConflictData
 
   constructor(data: ConflictData) {
-    super('VoyagerConflict')
+    super('GraphQLConflict')
     this.conflictInfo = data
   }
 }
--- a/packages/apollo-voyager-conflicts/src/api/ObjectConflictError.ts
+++ b/packages/apollo-voyager-conflicts/src/api/ObjectConflictError.ts
@@ -20,6 +20,11 @@ export interface ConflictData {
   clientData?: ObjectStateData
 
   /**
+   * Base data that before server side changes that caused conflict were applied
+   */
+  baseData?: ObjectStateData
+
+  /**
    * Flag used to inform client that conflict was already resolved on the server
    * and no further processing is needed. When flag is true `serverData` field will contain
    * resolved information. If value is false client will need to resolve conflict on their side
--- a/packages/apollo-voyager-conflicts/src/states/HashObjectState.ts
+++ b/packages/apollo-voyager-conflicts/src/states/HashObjectState.ts
@@ -1,13 +1,13 @@
-import * as debug from 'debug'
 import { ObjectState } from '../api/ObjectState'
 import { ObjectStateData } from '../api/ObjectStateData'
-import { CONFLICT_LOGGER } from '../constants'
+import { StatePersistence } from '../api/StatePersistence'
 
 /**
  * Object state manager using a hashing method provided by user
  */
 export class HashObjectState implements ObjectState {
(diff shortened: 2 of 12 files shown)


StephenCoady:
Just looked through this and it looks excellent! really easy to follow, you have thought of everything. ping me here when you are ready for verification and I will play with it!

packages/apollo-voyager-conflicts/src/states/HashObjectState.ts
```diff
@@ -1,27 +1,54 @@
...
+      this.statePersistence.persist(currentObjectState)
+    }
     // Hash can be calculated at any time and it is not added to object
     return currentObjectState
   }
+
+  public async previousState(currentObjectState: ObjectStateData) {
+    if (this.statePersistence) {
+      return await this.statePersistence.fetch(currentObjectState)
+    }
```
wtrocki: Undefined promise here

packages/apollo-voyager-conflicts/src/api/ObjectConflictError.ts
```diff
@@ -19,6 +19,11 @@ export interface ConflictData {
    */
   clientData?: ObjectStateData
 
+  /**
+   * Base data that before server side changes that caused conflict were applied
+   */
+  baseData?: ObjectStateData
```
wtrocki: Base data needs to be reflected in client interface (and be optional)

darahayes: Hey @wtrocki sorry if this is a silly question, but what exactly is the `baseData`? I'm not 100% sure.

Could you maybe describe an example where the base data is returned in the error and how it would be different from the server data and client data?

wtrocki: When resolving conflicts we usually have 3 types of artifacts:
- Input data (usually comes from client)
- Current source of truth data
- Base for both changes

There are numerous examples of this in real world. 
For example any type of merge client we tend to see 3 artifacts:
https://chodounsky.net/resources/img/posts/2013-05-09-merge-files-with-git-merge-tool-1.png

packages/apollo-voyager-conflicts/src/states/HashObjectState.ts
```diff
@@ -1,27 +1,54 @@
...
   constructor(hashImpl: (object: any) => string) {
     this.hash = hashImpl
   }
 
   public hasConflict(serverData: ObjectStateData, clientData: ObjectStateData) {
     if (this.hash(serverData) !== this.hash(clientData)) {
+      if (this.logger) {
+        this.logger.info(`Conflict when saving data.
+        current: ${serverData}, client: ${clientData}`)
+      }
```
wei-lee: is the `clientData` here is the one that before changes are made (currentState -1 ), or after the changes are made (currentState)? If it is actually the current state, doesn't it mean this will always cause a conflict?

wei-lee: or it depends on how the `hash` function is implemented: it needs to return a value that matches the current server data? Do you have an example of such a hash function?

wtrocki: Framework gives many options for integration. Currently we using VersionedObjectState (see VE aproach and we used it in example application in this repo and also on ionic example.
However integration tests are going to give us examples for entire api that is provided.

I identified some flaws in API here and going to get this fixed. 
Also working on diagram for this idea

wtrocki:
Note: This landed as an idea and we need more information in order to integrate that.

wtrocki:
Closing as outdated. We not going to save state persistence for the moment.

PR #48, opened 2019-01-09
feat: improve the conflict handling API

darahayes:
##### Checklist

- [ ] `npm test` passes
- [ ] tests are included


packages/apollo-voyager-conflicts/src/api/ObjectState.ts
```diff
@@ -24,18 +24,18 @@ export interface ObjectState {
...
   resolveOnClient(serverState: ObjectStateData, clientState: ObjectStateData): ConflictResolution
 
   /**
-  *
-  * @param serverState the current state of the object on the server
-  * @param clientState the state of the object the client wishes to perform some mutation with
-  * @param baseState the base object state that the client state is based off.
-  */
+   *
+   * @param serverState the current state of the object on the server
```
wtrocki: ❤️ naming here

darahayes:
Just working on getting some tests in there!

packages/apollo-voyager-conflicts/src/strategies/index.ts
```diff
@@ -0,0 +1,17 @@
+import { ConflictResolutionStrategy } from '../api/ConflictResolutionStrategy'
+import { ObjectStateData } from '../api/ObjectStateData'
+
+export interface StrategyMap {[key: string]: ConflictResolutionStrategy}
```
wtrocki: This seems strange to me. If we define object with set of strategies it will mean the same. We do not need interface for that.

darahayes: Fair enough, it was more so to strongly type the functions as ConflictResolutionStrategy types, however I can do that another way also

PR #50, opened 2019-01-09
feat: Enable pluggable logging

wtrocki:
Signed-off-by: Wojciech Trocki <wtrocki@redhat.com>
 
Pluggable logging :)

## Verification

1. Open conflict example app
2. Trigger conflict
3. See log messages

### DIFF
--- a/examples/conflicts/server.js
+++ b/examples/conflicts/server.js
@@ -5,6 +5,7 @@ const { ApolloVoyagerServer, gql } = require('../../packages/apollo-voyager-serv
 
 const { conflictHandler } = require('../../packages/apollo-voyager-conflicts')
 
+conflictHandler.enableLogging(console)
 
 // Types
 const typeDefs = gql`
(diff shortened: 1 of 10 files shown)


wtrocki:
@psturc WOW yuou are hero. Example apps have tests now: 
https://circleci.com/gh/aerogear/apollo-voyager-server/232?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Made me realize that we may stringify objects :)

examples/conflicts/server.js
```diff
@@ -5,6 +5,7 @@ const { ApolloVoyagerServer, gql } = require('../../packages/apollo-voyager-serv
 
 const { conflictHandler } = require('../../packages/apollo-voyager-conflicts')
 
+conflictHandler.enableLogging(console)
```
darahayes: I'm just wondering. conflictHandler (VersionedObjectState) is actually a class. It seems we never create an instance and are somehow statically accessing methods like `hasConflict`, `nextState` etc.

I think maybe we should rename the exported class to `ConflictHandler` and then we should be creating an instance like this:

```js
const conflictHandler = new ConflictHandler({ logger: console })
```

WDYT?

wtrocki: This is singleton created just for users  convenience. Sadly we cannot create interface constructor method. This way we are exposing default option without asking people to create object.

PR #57, opened 2019-01-10
fix: fix metrics middleware

darahayes:
Obviously, the bigger issue here is that our testing didn't catch this. With the rush to get this stuff ready for next week, I think we should solve this as part of a separate issue. https://github.com/aerogear/apollo-voyager-server/issues/58

- [x] `npm test` passes
- [x] tests are included

### DIFF
--- a/packages/apollo-voyager-metrics/src/index.ts
+++ b/packages/apollo-voyager-metrics/src/index.ts
@@ -82,7 +82,7 @@ export function updateResolverMetrics (resolverInfo: any, responseTime: number) 
 
 function applyMetricsMiddleware (app: Application, config: MetricsConfiguration) {
   const path = config && config.path ? config.path : '/metrics'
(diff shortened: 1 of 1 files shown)


packages/apollo-voyager-metrics/src/index.ts
```diff
@@ -82,7 +82,7 @@ export function updateResolverMetrics (resolverInfo: any, responseTime: number)
 
 function applyMetricsMiddleware (app: Application, config: MetricsConfiguration) {
   const path = config && config.path ? config.path : '/metrics'
-  app.get(path, app)
+  app.get(path, getMetrics)
```
wei-lee: haha, oh man...we should really have a test for this!!

darahayes: @wei-lee we will get it done see #58

PR #67, opened 2019-01-24
Doc: adding notes for how to bind to keycloak

secondsun:
### Description

Adding some basic instructions for how to inject keycloak config into a sync server app

### DIFF
--- a/doc/guides/add_keycloak_to_openshift.md
+++ b/doc/guides/add_keycloak_to_openshift.md
@@ -1,0 +1,57 @@
+## Prerequisites
+
+- There is a Keycloak service available.
+- You have provisioned a sync application using our playbook.
+- oc tools must be installed
+- You must first create a client for your application in the Keycloak Administration Console, and then click on the `Installation` tab, select `Keycloak OIDC JSON` for `Format` Option, and then click on `Download`. Save the downloaded `keycloak.json`.
+
+## Connecting your application to Keycloak
+
+Any application that connects to keycloak must consume a keycloak.json file.  This guide demonstrates how to add a keycloak.json file to your sync application deployment.  It will create the environment variable `KEYCLOAK_CONFIG` that contain the path to keycloak.json.  It is still your application's responsibility to consume this file.  We have provided an [example project](../../examples/keycloak).
+
+## Create a Keycloak Secret
+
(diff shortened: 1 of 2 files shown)


secondsun:
WIP, but feel free to offer general ideas.  It feels like it needs more prose, but really you just run a couple of commands and everythign "just works".

darahayes:
Nice content, however I think it might be unsuitable for the example docs which are designed to be quick, get off the ground locally scenarios. Could you create a separate doc for this under `doc/guides`? It will also help when I go to move this stuff over to `mobile-docs`

wtrocki:
I agree with @darahayes, but  Let's seed this inside mobile-doc repo instead. 
IMHO this repository should be just about sync and all connections to runtime will be added in mobile-docs repo, but no strong opinion about it.

secondsun:
@darahayes I've moved the changes to a separate file in guides.  @wtrocki I'll wait for dara to point me in the right place to throw this in mobile-docs.

doc/guides/add_keycloak_to_openshift.md
```diff
@@ -0,0 +1,57 @@
+## Prerequisites:
```
darahayes: Minor nit: no need for a colon here

doc/guides/add_keycloak_to_openshift.md
```diff
@@ -0,0 +1,57 @@
...
+- There is a Keycloak service available.
+- You have provisioned a sync application using our playbook.
+- oc tools must be installed
+- You must first create a client for your application in the Keycloak Administration Console, and then click on the `Installation` tab, select `Keycloak OIDC JSON` for `Format` Option, and then click on `Download`. Save the downloaded `keycloak.json`.
+
+## Connecting your application to Keycloak
+
+Any application that connects to keycloak must consume a keycloak.json file.  This guide demonstrates how to add a keycloak.json file to your sync application deployment.  It will create the environment variable `KEYCLOAK_CONFIG` that contain the path to keycloak.json.  It is still your application's responsibility to consume this file.  We have provided an [example project](../../examples/keycloak).
+
+## Create a Keycloak Secret
```
wtrocki: OpenShift template will help to simplify stuff a lot here, however we still do apb :(

PR #68, opened 2019-01-25
Added 3rd party webservices GraphQL integration example

wojta:
### Description
Example of integration with two 3rd party services.
The Star Wars API: https://swapi.co/
Namsor Gender Free API: https://api.namsor.com/

##### Checklist
- [x] tests are included and passes


examples/integrations/server.js
```diff
@@ -0,0 +1,141 @@
...
+
+// Resolver functions. This is our business logic
+const resolvers = {
+  Query: {
+    getGenderFromName,
+    getCharacterInfo
+  },
+  CharacterInfo: {
+    genderFromName: async (obj, args, context, info) =>
+      await getGenderFromName(
```
wojta: Not sure if it's the right approach to do the "sub-queries" in the resolvers. Some GraphQL expert might help...

examples/integrations/server.js
```diff
@@ -0,0 +1,141 @@
...
+        headers: {
+          "Content-Type": "application/json;charset=UTF-8"
+        }
+      }
+    );
+
+    const { data } = response;
+    return data;
+  } catch (e) {
+    console.error(e);
```
wtrocki: This is not proper behaviour as it will hang client indefinitely.

examples/integrations/server.js
```diff
@@ -0,0 +1,141 @@
...
+  type Query {
+    getCharacterInfo(id: ID!): CharacterInfo
+    getGenderFromName(name: String, surname: String): GenderInfo
+  }
+`;
+
+const getGenderFromName = async (obj, args, context, info) => {
+  try {
+    const { name, surname } = args;
+    const response = await axios.get(
```
wtrocki: With axios entire example becomes very trivial, however example seems to be overly big. 
I'm just looking for opinion if we really need demo for this as opposed to just documentation

examples/integrations/server.js
wtrocki: I would say remove this as purpose of this example is not related with sub-queries. It should be clearly show what it needs to show.

examples/test/integrations.test.js
```diff
@@ -0,0 +1,37 @@
...
+      scale
+      gender
+    }
+  }
+}
+`;
+
+test("Sending a query for Anakin Skywalker", async t => {
+  try {
+    const res = await sendQuery(query);
```
wtrocki: I'm worried if one of those apis will go down our integration tests will stop working. 
Maybe we could have just simple hello endpoint inside the example and make call for it?
This will make things more self contained and still show use cases.

wtrocki:
It looks good. I have added couple comments as suggestions. 
TL;DR - we can make this more simple because as someone completely new to this I will struggle to see what example is doing and where is actual integration/example happening.

doc/guides/examples.md
```diff
@@ -94,3 +94,16 @@ query hello {
 The query above will only work if the authenticated user has the `admin` role. You can see this rule being applied with the `@hasRole(role: "admin")` directive in [examples/keycloak/server.js](../../examples/keycloak/server.js#L22).
 
 Try change the the role to a made up role and restart the server. Try the sample query again and verify that an error is displayed.
+
+## 3rd party integrations example
+
+`integrations/index.js` shows, how it is easy to external web service services under a GraphQL based API.
```
StephenCoady: [Suggestion] `integrations/index.js` shows how easy it is to integrate the consumption of a REST API endpoint into your GraphQL resolver.

doc/guides/examples.md
```diff
@@ -94,3 +94,16 @@ query hello {
...
+
+```
+$ node examples/integrations/index.js
+🚀 Server ready at http://localhost:4000/graphql
+```
+
+Open http://localhost:4000/graphql and you can run already prepared query there.
+
+Columns that are directly in `getCharacterInfo` are resolved from [Star Wars People API](https://swapi.co/).
```
StephenCoady: @wojta since most of the integration will be in the resolvers do you think it might make sense to document a simple code example here? Something very easy like:
The suggested pattern for integrating with a REST API is a simple flow as follows:
```
const resolvers = {
  Query: {
    queryREST: async (obj, args, context) => {
       const result = axios.get(YOUR_ENDPOINT_HERE, params);
       return result;
    }
  }
}
```
for a working example, please see {FILE} etc...

Not sure about using a library name like axios in the docs but we could just say you can use any http library you like.

doc/guides/examples.md
```diff
@@ -94,3 +94,16 @@ query hello {
 The query above will only work if the authenticated user has the `admin` role. You can see this rule being applied with the `@hasRole(role: "admin")` directive in [examples/keycloak/server.js](../../examples/keycloak/server.js#L22).
 
 Try change the the role to a made up role and restart the server. Try the sample query again and verify that an error is displayed.
+
+## 3rd party integrations example
```
wtrocki: It's hard to get what this does. Maybe:

"Integrating with external RESTfull API"?

examples/integrations/server.js
```diff
@@ -0,0 +1,74 @@
...
+// Types
+const typeDefs = gql`
+  type CharacterInfo {
+    name: String
+    homeworld: String
+    gender: String
+  }
+  type Query {
+    getCharacterInfo(id: ID!): CharacterInfo
+    hello: String
```
wtrocki: Remove hello

examples/integrations/server.js
```diff
@@ -0,0 +1,74 @@
...
+  }
+`;
+
+// Resolver functions. This is our business logic
+const resolvers = {
+  Query: {
+    getCharacterInfo: async (obj, args, context, info) => {
+      //gets JSON response from the external API, parameter `id` is passed to the external API
+      const response = await axios.get(
+        `http://swapi.co/api/people/${args.id}`,
```
wtrocki: Missing default value can cause error. 
`https://swapi.co/api/people/undefined/` Add random fela to avoid this.

examples/test/basic.test.js
```diff
@@ -1,27 +1,19 @@
-const test = require('ava')
-const axios = require('axios')
-const { server, app } = require('../basic/server')
+const test = require("ava");
```
wtrocki: Let's not start holly war. If there were no style changes in the file let's revert this changes.

.vscode/launch.json
```diff
@@ -46,6 +45,12 @@
       "outFiles": [
         "${workspaceFolder}/**/*.js"
       ]
+    },
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "Launch Integrations Example",
+      "program": "${workspaceFolder}/examples/integrations/index.js"
```
wtrocki: I will suggest to rename this example from `integrations` to `rest` or someting like that. 
Integrations are too generic thing

examples/test/integrations.test.js
```diff
@@ -0,0 +1,17 @@
+const test = require("ava");
+const { server, app } = require("../integrations/server");
+const port = 4000;
+app.listen({ port });
+const sendQuery = query =>
+  require("./util/sendQuery")(query, server.graphqlPath, port);
+
+test("Sending a request to basic example app should return OK status code and no errors", async t => {
```
wtrocki: Looks like this is unfinished? 
Testing hello only?

doc/guides/examples.md
wojta: I'll take any suggestion. My English is pretty bad. :)

doc/guides/examples.md
wtrocki: Yes. Yes. Yes Let's not mention API we integrate here and build generic example.

examples/test/basic.test.js
wojta: I just added `sendQuery` as shared `util/sendQuery.js` to shorten this test.

examples/test/integrations.test.js
wtrocki: Ah.. But then this kind of test will tell us if API is down. 
I was like - let's use internal REST API inside app, but if we use external API anyway let's test it and remove hello.

doc/guides/examples.md
wojta: Ok, I did it somehow. Linked example's source lines.

examples/integrations/server.js
wojta: removed

wojta:
@wtrocki Fixed it, is this ok?

wtrocki:
Rebase needed. @darahayes WDYT?

darahayes:
Hey @wojta thanks a lot for this PR. Coming to think of it, I'm a little bit unsure about the overall direction of this PR for a couple of reasons. The code and everything is good but I'm not sure about the API choice.

* I'm a massive star wars fan but I feel like this API is not really suitable. From a cultural perspective, star wars is not something everybody likes or knows about, and could be a possible barrier to entry for some people.
* At the end of the day, our goal is to build a commercial OSS project that we can present in the best possible light. I think it's best if we avoid any potentially sensitive topics in our code/docs and also avoid anything that may come across as unprofessional.
* On a purely technical note, I'm happy enough if we use an external API but our tests will fail if that external API is down so my suggestion is that we can ensure the test does not fail when it's down.

I was thinking we could maybe use this vehicles API instead.

```
https://vpic.nhtsa.dot.gov/api/vehicles/getmodelsformake/volkswagen?format=json
```

the results look like this:

```
{"Count":41,"Message":"Response returned successfully","SearchCriteria":"Make:volkswagen","Results":[{"Make_ID":482,"Make_Name":"Volkswagen","Model_ID":1951,"Model_Name":"Routan"},
{"Make_ID":482,"Make_Name":"Volkswagen","Model_ID":3133,"Model_Name":"Golf"},
{"Make_ID":482,"Make_Name":"Volkswagen","Model_ID":3134,"Model_Name":"Passat"},
{"Make_ID":482,"Make_Name":"Volkswagen","Model_ID":3135,"Model_Name":"Phaeton"},
{"Make_ID":482,"Make_Name":"Volkswagen","Model_ID":3136,"Model_Name":"Touareg"},
{"Make_ID":482,"Make_Name":"Volkswagen","Model_ID":3137,"Model_Name":"Jetta"},
{"Make_ID":482,"Make_Name":"Volkswagen","Model_ID":3140,"Model_Name":"GTI"},
{"Make_ID":482,"Make_Name":"Volkswagen","Model_ID":3141,"Model_Name":"R32"},
{"Make_ID":482,"Make_Name":"Volkswagen","Model_ID":3142,"Model_Name":"Jetta Wagon"},
{"Make_ID":482,"Make_Name":"Volkswagen","Model_ID":4084,"Model_Name":"Rabbit"}]}
```

you can swap out the name `volkswagen` with anything else

Our schema could be something like.

```gql
type Vehicle: {
  manufacturer: String!
  model: String!
}

type Query {
  getModelsForManufacturer(manufacturer: String!): [Vehicles]
}
```

then in our business logic,

```js
{ 
  getModelsForManufacturer: (obj, info, args, context) => {
    const response = await axios.get(`https://vpic.nhtsa.dot.gov/api/vehicles/getmodelsformake/${args.manufacturer}?format=json`)
    const result = response.Results.map((item) => { return { manufacturer: item.Make_Name, model: item.Model_Name }})
    return result
  }
}
```

What do you think about that idea?

wojta:
I agree with @darahayes. But still use 3rd party API, we can't be sure even about our own server running.

PR #71, opened 2019-01-25
fix: Add documentation for conflict example

wtrocki:
# Motivation

Add documentation for conflict example
With such big amount of examples it probably makes sense to split them into separate files or move to example folder?

### DIFF
--- a/doc/guides/examples.md
+++ b/doc/guides/examples.md
@@ -97,16 +97,16 @@ Try change the the role to a made up role and restart the server. Try the sample
 
 ## Conflict Example
 
-`conflict/index.js` provides example for server side implementation of data conflict package that can be used with conjunction with `apollo-voyager-server` framework to detect and handle data conflicts.
+This example demonstrates how to use the `@aerogear/vogayer-conflicts` package to detect and handle data conflicts within the resolver functions.
 
-### Running example
+### Running Example
 
 ```
 $ node examples/conflicts/index.js
 🚀 Server ready at http://localhost:4000/graphql
 ```
 
-### Testing conflict resolution
+### Testing Conflict Resolution
 
 Open [http://localhost:4000/graphql](http://localhost:4000/graphql).
 You will see the GraphQL Playground. This is a space where you can try out queries and see the results.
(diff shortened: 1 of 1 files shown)


wtrocki:
It could be nice to sort out some state with example apps by having docs close to them or in root folder.

wtrocki:
Created as separate PR but it may be strictly related to #72

wtrocki:
Ready for review

wtrocki:
@darahayes How about we reformat examples to something like:

```
REAME.md -> There are examples here:
Examples/REadme.md -> List of examples with on sentence description
Individual folders -> Current content
```
This will put documentation close to examples so it will be discoverable by people

darahayes:
Hey @wtrocki initially, I put the examples in a single file because I knew I/we were much more likely to maintain it. It's also very easy to point to one file and say, 'All of the examples are in here, knock yourself out.' As a reader you don't need to jump from file to file. You can just follow the one page and go through all of the examples tutorial style which I thought was nice. (maybe a person is less likely to try all the examples if they need to jump to all these different documents)

That being said, I think your suggestion is a good way moving forward. Feel free to bring it in whichever PR you like.

wtrocki:
Review anyone? I'm going to apply suggestion in separate PR.

doc/guides/examples.md
```diff
@@ -94,3 +94,38 @@ query hello {
 The query above will only work if the authenticated user has the `admin` role. You can see this rule being applied with the `@hasRole(role: "admin")` directive in [examples/keycloak/server.js](../../examples/keycloak/server.js#L22).
 
 Try change the the role to a made up role and restart the server. Try the sample query again and verify that an error is displayed.
+
+## Conflict Example
+
+`conflict/index.js` provides example for server side implementation of data conflict package that can be used with conjunction with `apollo-voyager-server` framework to detect and handle data conflicts.
```
darahayes: [Suggestion]: This could be reworded to something like:

```markdown
This example demonstrates/shows how to use the `@aerogear/vogayer-conflicts` package to detect and handle data conflicts within the resolver functions.
```

PR #73, opened 2019-01-28
fix:add check for subscriptions when wrapping resolvers

StephenCoady:
Fixes: https://issues.jboss.org/browse/AEROGEAR-8438

### DIFF
--- a/packages/apollo-voyager-tools/package.json
+++ b/packages/apollo-voyager-tools/package.json
@@ -21,8 +21,6 @@
     "graphql-tools": "^4.0.3"
   },
   "devDependencies": {
-    "@aerogear/apollo-voyager-audit": "0.2.7-rc3",
-    "@aerogear/apollo-voyager-metrics": "0.2.7-rc3",
     "@aerogear/apollo-voyager-server": "0.2.7-rc3",
     "ava": "1.0.0-rc.2",
     "ts-node": "^7.0.1",
--- a/packages/apollo-voyager-tools/package.json
+++ b/packages/apollo-voyager-tools/package.json
@@ -21,6 +21,9 @@
     "graphql-tools": "^4.0.3"
   },
   "devDependencies": {
+    "@aerogear/apollo-voyager-audit": "0.2.7-rc3",
+    "@aerogear/apollo-voyager-metrics": "0.2.7-rc3",
+    "@aerogear/apollo-voyager-server": "0.2.7-rc3",
     "ava": "1.0.0-rc.2",
     "ts-node": "^7.0.1",
     "typescript": "^3.1.6"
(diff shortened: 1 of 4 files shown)


wtrocki:
Unit test will be handy for this case,  but please do not feel that it is required.

darahayes:
+1 some unit tests here would be really nice.

* The happy case with no subscriptions.
* The case with a mixture of subscriptions and regular resolvers

This is a core function which has 0 code coverage right now. I know it will take you some extra time @StephenCoady but it would be a massive help!

wtrocki:
@StephenCoady Rebase on top of master and force push

packages/apollo-voyager-tools/src/wrapResolvers.test.ts
```diff
@@ -0,0 +1,59 @@
...
+    return (obj, args, context, info) => {
+      return context
+    }
+  }
+}
+
+const resolversWithOutSubscription = {
+  Query: {
+    allTasks: async (context: any) => {
+      return context.db.select().from('tasks')
```
darahayes: It's kinda weird to see this code in our unit tests. Let's have a chat about how to test this stuff properly

packages/apollo-voyager-tools/src/wrapResolvers.test.ts
```diff
@@ -0,0 +1,59 @@
...
+  }
+}
+
+test('Without subscriptions', (t) => {
+  const config: any = {
+    auditLogger,
+    metrics,
+    securityService: {} as any
+  }
+  const result = wrapResolvers(voyagerResolvers(resolversWithOutSubscription, config), resolverFunction())
```
darahayes: So ideally, we want to test that our resolver functions were correctly wrapped with the wrapper function provided and that when we now execute our resolver, both the wrapper and the original resolver were both called. Can you think of any ideas about how we might do that? 🙂

packages/apollo-voyager-tools/package.json
```diff
@@ -21,6 +21,9 @@
     "graphql-tools": "^4.0.3"
   },
   "devDependencies": {
+    "@aerogear/apollo-voyager-audit": "0.2.7-rc3",
```
darahayes: Just a minor thing: make sure this and the metrics ones are removed.

StephenCoady: ahh, never saved the file. 2 secs

packages/apollo-voyager-tools/package.json
```diff
@@ -21,6 +21,7 @@
     "graphql-tools": "^4.0.3"
   },
   "devDependencies": {
+    "@aerogear/apollo-voyager-server": "0.2.7-rc3",
```
wtrocki: This is circular dependency. No way this is going to work. 
We got lucky with the build.
Disabled that and logged separate issue #80

darahayes: @wtrocki this is only a dev dependency used in the tests to pull in some interface needed to allow compilation. It's not required in the production dependencies.

wtrocki: Doesn't matter - when you will end up with clean project without stuff compiled typescript compiler will be forced to resolve `chicken or egg problem`.

PR #76, opened 2019-01-29
Cleanup Keycloak example realm

danielpassos:
It makes the `realm-export.json` a little more easy to read and provides the same behavior of the previous one


wtrocki:
This realm-export is used internally for tests and for example. 
For us it's just easier to maintain it by dropping entire realm from keycloak. 
It may have some noise but no one is checking that anway. 
Users will normally get realm and all config thru binding in MDC so this is more for example purposes.

PR #77, opened 2019-01-29
BREAKING: Drop apollo from package names

wtrocki:
Drop apollo from package names and remove references in some places.


darahayes:
@wtrocki +1 on `0.3.0`

doc/api/apollo-voyager-server.md
```diff
@@ -1,6 +1,6 @@
-# apollo-voyager-server API
```
darahayes: This file should probably be renamed `voyager-server.md`

darahayes:
@wtrocki should we also rename the folders under the `packages` directory?

PR #78, opened 2019-01-29
More tight context integration (including conflict in context)

wtrocki:
# Motivation

Packages can be self sufficient if we move basic types inside them.
This is going to give us couple major advantages:

- Packages are decoupled from the server and can be used with plain Apollo if needed
- Each package contains default implementation that we support
- Single responsibility in one place
- Better aproach for documenting
- Simple dependency model (server requires individual packages)
- Makes server only and integration/getting started experience package
- It makes things simply to understand and contribute

##### Checklist

- [x] Reactor other packages to follow the same aproach as conflict
- [x] Add type support for context
- [x] Added conflict to context
- [x] Fixed broken types for audit
- [x] Adjusted example app
- [x] Fixed broken master


wtrocki:
Just lost all the changes due to wrong rebase :(

wtrocki:
#RebaseNightmare

wtrocki:
Horrible rebase. Closing

PR #81, opened 2019-01-29
More tight context integration (including conflict in context)

wtrocki:
# Motivation

Add conflict to contex and align architecture between packages.
Packages can be self sufficient if we move basic types to them from server package.
This is going to give us couple major advantages:

- Packages are decoupled from the server and can be used with plain Apollo if needed
- Each package contains default implementation that we support
- Single responsibility in one place
- Better aproach for documenting
- Simple dependency model (server requires individual packages)
- Makes server only and integration/getting started experience package
- It makes things simply to understand and contribute

##### Checklist

- [x] Reactor other packages to follow the same aproach as conflict
- [x] Add type support for context
- [x] Added conflict to context
- [x] Fixed broken types for audit
- [x] Adjusted example app
- [x] Fixed broken master


wtrocki:
Ready for review.

packages/voyager-keycloak/src/api/AuthContextProvider.ts
```diff
@@ -0,0 +1,5 @@
+
+export interface AuthContextProvider {
+  isAuthenticated (): boolean
+  hasRole (role: string): boolean
```
wtrocki: Created issue to add docs for this

wtrocki: #82

darahayes:
@wtrocki reviewing now! Can you remove the top level `package-lock.json`? Makes it hard to gauge the impact of the PR in terms of lines added/deleted.

wtrocki:
@darahayes Yes. I overlooked that.

packages/voyager-server/package.json
```diff
@@ -22,6 +22,8 @@
   },
   "dependencies": {
     "@aerogear/voyager-audit": "^0.3.0",
+    "@aerogear/voyager-conflicts": "^0.3.0",
```
darahayes: Is this necessary inside actual deps or just dev deps?

wtrocki: Idea is that all will be inside actual production dependencies as others.

PR #84, opened 2019-01-30
feat: Examples docker container improved publishing

wojta:
### Description
Improved publishing

### Checklist
- [x] go to `examples`
- [x] launch npm run push-basic
- [x] successfully publishes current version to the Dockerhub

### Alternative
also possible to publish by script directly
```sh
$ ./push-script.sh latest basic
```
First parameter is container's TAG and second is the directory with the example.


examples/basic/server.js
```diff
@@ -1,6 +1,6 @@
 const express = require('express')
-
-const { ApolloVoyagerServer, gql } = require('../../packages/apollo-voyager-server')
+const { LOCAL } = process.env; //use this environment variable to run with code from local repository not the published version
+const { ApolloVoyagerServer, gql } = LOCAL ? require('../../packages/apollo-voyager-server') : require("@aerogear/apollo-voyager-server");
```
wtrocki: Ah.. We do not need that anymore. `"@aerogear/apollo-voyager-server"` will work and we have this on master.

examples/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM node:8
+WORKDIR /usr/src/app
+COPY package*.json ./
+COPY $directory $directory
+RUN npm install
+EXPOSE 4000
+CMD [ "node", "$directory" ]
```
wei-lee: this is not a valid dockerfile, it probably better rename it to be something like `Dockerfile.template`. Otherwise people may try to run `docker build` from this dir

examples/package.json
```diff
@@ -1,13 +1,15 @@
...
   "private": true,
   "description": "",
   "scripts": {
     "compile": "exit 0",
     "watch": "exit 0",
     "compile:clean": "exit 0",
-    "test-examples": "ava --concurrency=1 test/*.test.js"
+    "test-examples": "ava --concurrency=1 test/*.test.js",
+    "push-basic": "./push-script.sh $npm_package_version basic",
+    "push-keycloak": "./push-script.sh $npm_package_version keycloak"
```
wei-lee: sorry for my lack of knowledge, is `npm_package_version` set by npm automatically?

examples/push-script.sh
```diff
@@ -0,0 +1,18 @@
...
+echo "Publishing version $1 of voyager-server examples to Docker"
+
+i=1
+for dir in "$@"
+do
+  if [ "$i" -gt 1 ]; then
+      echo "Building docker for example $dir"
+      DOCKERFILE="Dockerfile.$dir"
+      CONTAINER="aerogear/voyager-server-example-$dir:$1"
+      cat Dockerfile | sed -e "s/\$directory/$dir/g" > "$DOCKERFILE"
```
darahayes: This approach of using sed to replace the `$directory` string is kinda confusing. Dockerfile as support for something called build args which should probably be used instead. https://docs.docker.com/engine/reference/builder/#arg

Basically in the dockerfile you would make a declaration `ARG directory`

and then when you build you can do `docker build --build-arg directory=$DIR ...`

If you use a build arg, this script will become much simpler as you will not need to make copies of the Dockerfile for each example you are building 🙂

examples/push-script.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/sh
+#script takes first param as current Voyager release version and next params the names of directories of examples that should be pushed
```
darahayes: Instead of taking the current release version as an argument, it would be nicer to take it as an environment variable. See this example here: https://github.com/aerogear/voyager-server/blob/master/scripts/validateRelease.sh#L3-L4

This keeps the way in which we invoke our bash scripts consistent which is especially important for CI related stuff. It's also a bit more readable to have nicely named variables IMO

The end result will be instead of calling the script with `./push-script.sh $package_version basic` we will do `TAG=$package_version ./push-script.sh basic`

examples/package.json
wojta: yes it is.

examples/push-script.sh
wojta: Ok, my lack of knowledge. I'll fix it to do that...

examples/push-script.sh
wojta: ok

wojta:
Thanks for your inputs, I'll change it and update soon.

examples/Dockerfile
wojta: using args now

examples/basic/server.js
wojta: removed

PR #86, opened 2019-01-31
feat: CircleCI pushes basic example server container on release

wojta:
https://issues.jboss.org/browse/AEROGEAR-8436

### Description

Publishes `aerogear/voyager-server-example-basic` after each release (tag) and on master builds.

Tested in my fork.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -74,7 +74,10 @@ jobs:
       - setup_remote_docker:
           version: 17.05.0-ce
           docker_layer_caching: true
-      - run: cd examples && TAG=$CIRCLE_TAG ./push-script.sh basic
+      - run: |
+          cd examples
+          docker login -u $DOCKERHUB_USERNAME -p $DOCKERHUB_PASSWORD
+          TAG=$CIRCLE_TAG ./push-script.sh basic
 workflows:
   version: 2
   build_and_release:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -65,13 +65,16 @@ jobs:
       - run: TAG=$CIRCLE_TAG npm run release:validate
       - run: TAG=$CIRCLE_TAG npm run publish
   containers_publish:
-    working_directory: ~/examples
+    working_directory: ~/aerogear
     docker:
       # Node 8 LTS
-      - image: circleci/node:carbon
+      - image: docker:17.05.0-ce-git
     steps:
       - checkout
-      - run: TAG=$CIRCLE_TAG ./push-script.sh basic
+      - setup_remote_docker:
+          version: 17.05.0-ce
+          docker_layer_caching: true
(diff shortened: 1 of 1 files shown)


wojta:
@wtrocki How can I check if this really works?

.circleci/config.yml
```diff
@@ -64,6 +64,7 @@ jobs:
       - run: CI=true npm run release:prep
       - run: TAG=$CIRCLE_TAG npm run release:validate
       - run: TAG=$CIRCLE_TAG npm run publish
+      - run: cd examples && TAG=$CIRCLE_TAG ./push-script.sh basic
```
wei-lee: -1 for having this as part of `npm_publish`. It should have its own task.

wojta: moved as separate job

wojta:
Ok, this won't work at all. There is still a hack to be done. Let me fix it.
https://circleci.com/docs/2.0/building-docker-images/

wojta:
It's now fixed. I tried to run this workflow in my CircleCI - https://circleci.com/gh/wojta/apollo-voyager-server/31#build-parameters/containers/0

wtrocki:
Do not want to spoil the party but we pick pretty hard way to implement that. 
Maybe we can use something official from Circle CI that will not require maintenance and testing.

Maybe something like this: https://circleci.com/orbs/registry/orb/circleci/docker-publish

wtrocki:
Actually if that works then I do not mind really.   Leaving approval to @darahayes as keeps eye on our automation standards across the repos.

wojta:
Actually it seems both as official way (see link to docs above). I would prefer to leave it the way it is.

.circleci/config.yml
```diff
@@ -64,7 +64,21 @@ jobs:
       - run: CI=true npm run release:prep
       - run: TAG=$CIRCLE_TAG npm run release:validate
       - run: TAG=$CIRCLE_TAG npm run publish
-
+  containers_publish:
```
darahayes: Let's rename this to `publish_example_containers`

.circleci/config.yml
```diff
@@ -85,4 +99,13 @@ workflows:
             tags:
               only: /.*/ # allow anything because tag syntax is validated as part of validate-release.sh
             branches:
-              ignore: /.*/
\ No newline at end of file
+              ignore: /.*/
+      - containers_publish:
+          requires:
+            - test_examples
```
darahayes: I think this should also require `- npm_publish` considering that the new images will need to have the new packages published to the registry before we can install them as part of the container build. We may need to test that one out because sometimes it takes some time before published packages become available on the registry.

wojta: done

.circleci/config.yml
wojta: done

PR #87, opened 2019-02-01
AEROGEAR-8413 Audit logs for conflicts

aliok:
# Verification:
- Link stuff
```
npm run bootstrap
npm run compile
lerna exec -- npm link .
cd examples
npm link @aerogear/voyager-tools
npm link @aerogear/voyager-audit
npm link @aerogear/voyager-conflicts
npm link @aerogear/voyager-keycloak
npm link @aerogear/voyager-metrics
npm link @aerogear/voyager-server
```
- Run conflict example
- http://localhost:4000/graphql
- End up in conflict
- See the console output

@wtrocki @darahayes 
I took the approach to make things working as we talked in our last meeting.
I won't do any refactoring part of this PR.

### DIFF
--- a/examples/conflicts/server.js
+++ b/examples/conflicts/server.js
@@ -2,6 +2,7 @@ const express = require('express')
 const queries = require('./queries')
 const { VoyagerServer, gql } = require('@aerogear/voyager-server')
 const metrics = require('@aerogear/voyager-metrics')
+const auditLogger = require('@aerogear/voyager-audit')
 
 // Types
 const typeDefs = gql`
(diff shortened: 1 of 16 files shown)


aliok:
@wtrocki @darahayes  ready for review

examples/conflicts/server.js
```diff
@@ -54,8 +56,9 @@ const resolvers = {
       greeting = conflict.nextState(args)
       return greeting
     },
-    changeGreetingClient: async (obj, args, { conflict }, info) => {
-      if (conflict.hasConflict(greeting, args, info)) {
+    changeGreetingClient: async (obj, args, context, info) => {
+      const { conflict } = context
+      if (conflict.hasConflict(greeting, args, obj, args, context, info)) {
```
wtrocki: IMHO it will be cool to have separate function here for audit that can be used to pass that arguments. 
Alternatively we can mount all of those into context before and just pass only context. 
Alternatively we could aggregate them into single object - audit log.

Audit log is an additional feature here but it obfuscates conflict API so we may have couple quick wins.
Note: I'm not saying that we should do that now, but let's have discussion on this while we have PR opened.

wtrocki:
I have tried that and it works. We could try to get quick win on conflict API, but I have no strong opinion to do so.

examples/conflicts/server.js
aliok: can you write some pseudo code? I didn't understand

wtrocki:
@aliok Feel free to improve that. Verified the changes so approving this PR.

PR #88, opened 2019-02-01
feat: restapi example own exception handling and throwing

wojta:
### Description
https://issues.jboss.org/browse/AEROGEAR-8460

Improved error/exception handling in the `restapi` example.

### DIFF
--- a/examples/restapi/README.md
+++ b/examples/restapi/README.md
@@ -6,11 +6,31 @@ Using GraphQL it's possible to write resolvers that call some other APIs.
 const resolvers = {
   Query: {
     queryREST: async (obj, args, context) => {
-       const result = await axios.get(YOUR_ENDPOINT_HERE, params);
-       return result.data;
+      try {
+        const result = await axios.get(YOUR_ENDPOINT_HERE, params);
+        return result.data;
+      } catch (e) {
+        throw new Error("some error");
+      }
     }
   }
-}
+};
+```
+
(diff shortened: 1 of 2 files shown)


examples/restapi/README.md
```diff
@@ -6,11 +6,31 @@ Using GraphQL it's possible to write resolvers that call some other APIs.
 const resolvers = {
   Query: {
     queryREST: async (obj, args, context) => {
-       const result = await axios.get(YOUR_ENDPOINT_HERE, params);
-       return result.data;
+      try {
+        const result = await axios.get(YOUR_ENDPOINT_HERE, params);
+        return result.data;
+      } catch (e) {
+        throw new Error("some error");
```
wtrocki: GraphQLError ?

examples/restapi/README.md
```diff
@@ -6,11 +6,31 @@ Using GraphQL it's possible to write resolvers that call some other APIs.
...
+      }
     }
   }
-}
+};
+```
+
+It can happen that external API is down or some other error occured. If you handle the error, you can throw your own exception and it will get passed to client as an `errors` field.
+
+```js
```
wtrocki: IMHO this is redundant but we may keep that.

examples/restapi/README.md
wojta: I didn't mention any difference here, but ok.

wtrocki: If we provide info how we handle errors we should do it the way that GraphQL recommends that. 
It's small thing but it will make huge difference.

PR #95, opened 2019-02-06
feat: Generate API using tsdocs

wtrocki:
Generate api documentation using script integrated with current docs.

## Verification

Execute provided script and check if API is generated properly.


darahayes:
@wtrocki could you delete the package-lock.json and add a `.gitignore` entry for it? :)

wtrocki:
Pinging for review?

PR #96, opened 2019-02-06
fix: WIP idea to fix Audit logging API

wtrocki:
Current api is very hard to document and explain. 
This simple trick will do the job. 
Posting as idea.


wtrocki:
Going to rebase this after anyway.

PR #102, opened 2019-02-08
breaking: several improvements and do not enable /token endpoint by default

darahayes:
So I'm thinking about writing a short blog post about using `voyager-keycloak` in plain Apollo as I think this module has potential value outside of Voyager, in the Keycloak community and potentially the GraphQL community.
In the process of using voyager-keycloak with plain Apollo I found some issues which I decided to fix.

Summary of changes.

* breaking: `/token` endpoint is disabled by default. This is important for production usage. It is possible to enable it explicitly with `service.applyAuthMiddleware(expressRouter, { tokenEndpoint: true })` See the example app.
* originally `hasRole` had some code that like `const token = context.request.kauth....` which meant the `request` object had to be included inside the GraphQL context. This is totally unnecessary because you can get `context.auth.accessToken`. This is not usually a problem because `voyager-server` automatically adds this, but if you want to use `voyager-keycloak` with plain Apollo then it means the developer had to add extra pointless stuff to the context which is bad.
* Ability to pass a logger into the `KeycloakSecurityService` at a top level. Note I haven't figured out how to get a logger down into the hasRole code yet. @wtrocki perhaps we could figure that one out together.
* Tonnes of improvements to the `SecurityService` interface. Much more strongly typed. Still not perfect but way better!

- [x] `npm test` passes
- [x] tests are included


wtrocki:
Great idea for a blog. Happy to collaborate on this.

packages/voyager-keycloak/package.json
```diff
@@ -21,9 +21,9 @@
     "test": "ava -v '*.test.ts' '**/*.test.ts'"
   },
   "dependencies": {
-    "@aerogear/voyager-server": "^0.4.0-rc2",
```
wtrocki: wow.. I need to check if other packaces have this as well. Good catch.

packages/voyager-keycloak/src/AuthContextProvider.ts
```diff
@@ -13,7 +11,7 @@ export class KeycloakAuthContextProvider implements AuthContextProvider {
     this.authenticated = !!((req && req.kauth && req.kauth.grant) ? req.kauth.grant.access_token : undefined)
   }
 
-  public isAuthenticated () {
+  public isAuthenticated (): boolean {
```
wtrocki: I haven't checked that before but this may be bug. Authenticated is basing on flag that is set once. User can be logged out etc. but authenticated will still return true.

packages/voyager-keycloak/src/KeycloakSecurityService.ts
```diff
@@ -64,13 +72,15 @@ export class KeycloakSecurityService implements SecurityService {
...
 
-    expressRouter.get('/token', keycloak.protect(), function (req, res) {
-      if (req.session && req.session['keycloak-token']) {
-        return res.json({
-          'Authorization': 'Bearer ' + JSON.parse(req.session['keycloak-token']).access_token
-        })
-      }
-      res.json({})
-    })
+    if (tokenEndpoint) {
```
wtrocki: Maybe we could jusf add that to ionic app backend. Flag is resolving issue

packages/voyager-keycloak/src/AuthContextProvider.ts
darahayes: For the duration of a request the user is authenticated right? You can't log out in the middle of a request

darahayes: How it works is the AuthContextProvider is created on every incoming request. This `isAuthenticated` here is really just saying 'is there a valid token on the request?' Seeing as you cannot really log out in the middle of a request I don't think this is a bug. What are your thoughts on that?

wtrocki: Ah.. I got impression that this is created once.

wtrocki:
Good to merge!

PR #105, opened 2019-02-08
feat: added new methods to handle tokens

StephenCoady:
add ability to manipulate tokens in the same way keycloak do. this is not ideal but they do not expose this logic by default. 
related to: 

https://github.com/aerogear/aerogear-js-sdk/pull/240
https://github.com/aerogear/ionic-showcase/pull/50

### DIFF
--- a/packages/voyager-keycloak/src/KeycloakSecurityService.ts
+++ b/packages/voyager-keycloak/src/KeycloakSecurityService.ts
@@ -11,6 +11,7 @@ export class KeycloakSecurityService implements SecurityService {
   public readonly schemaDirectives: any
   public readonly authContextProvider: any
   public readonly log: any
+  public keycloak: any
 
   constructor (keycloakConfig: any) {
     this.keycloakConfig = keycloakConfig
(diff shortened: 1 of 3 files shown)


packages/voyager-keycloak/src/KeycloakSecurityService.ts
```diff
@@ -73,4 +74,41 @@ export class KeycloakSecurityService implements SecurityService {
...
+    const tokenObject = this.getTokenObject(token)
+    const result = await this.keycloak.grantManager.validateToken(tokenObject, 'Bearer')
+    return (result === tokenObject) ? true : false
+  }
+
+  public getTokenObject(token: string): KeycloakToken {
+    let fullToken: KeycloakToken = {}
+    if (token) {
+      try {
+        const parts = token.split('.')
```
wtrocki: If this is copy paste I will put that somewhere outside this file and link to original code 
so we could have simple way to update it in case of potential problems etc.

packages/voyager-keycloak/src/KeycloakSecurityService.ts
```diff
@@ -73,4 +74,41 @@ export class KeycloakSecurityService implements SecurityService {
       res.json({})
     })
   }
+
+  public async validateToken(token: string): Promise<boolean> {
+    const tokenObject = this.getTokenObject(token)
+    const result = await this.keycloak.grantManager.validateToken(tokenObject, 'Bearer')
+    return (result === tokenObject) ? true : false
```
wtrocki: :)

wtrocki:
REbase needed.

StephenCoady:
@wtrocki good to go

packages/voyager-keycloak/src/KeycloakToken.ts
```diff
@@ -0,0 +1,34 @@
+// This function is taken directly from Keycloak's connector
+// here: https://github.com/keycloak/keycloak-nodejs-connect/blob/master/middleware/auth-utils/token.js#L30-L47
```
wtrocki: Best to use commits instead of master as master will move and lines will change.

wtrocki: Ignore that, but leaving just FYI

packages/voyager-keycloak/src/schemaDirectives/hasRole.ts
```diff
@@ -24,7 +24,7 @@ export class HasRoleDirective extends SchemaDirectiveVisitor {
   public visitFieldDefinition (field: any) {
     const { resolve = defaultFieldResolver } = field
     const { error, value } = this.validateArgs()
-
+    // TODO ignore for subscriptions
```
wtrocki: ?? :)

StephenCoady: i think dara has another pr that actually escapes this, need to check and I will remove this comment

PR #109, opened 2019-02-13
AEROGEAR-8514 - Collect client info in sync server side and expose th…

aliok:
…em in the metrics endpoint

Modifications:
1. Collect unique client and user count and give that info to Prometheus
2. Add success and authenticated labels to 2 metrics where it makes sense, so that we can create a success/failure and authenticated/anonymous graphs in Grafana
3. We didn't record the failed resolvers for Prometheus. Now with the introduction of success/failure above, we do it.

@psturc testing this whole thing is a pain in the ass. I have an [example application](https://github.com/aliok/nodejs-ex) which uses Voyager framework with metrics+auditLogs+keycloak and I have a [tool](https://github.com/aliok/aerogear-sync-metrics-generator/) that executes lots of GraphQL requests against that server.
But you need to do `npm link`, Keycloak configs and things like that. 

If we don't want to go that way here's a smoke test:

* Run metrics example and go to http://localhost:4000/graphql
* Execute `{hello}` bunch of times
* See these metrics (they're new)
```
unique_clients 1
unique_users 1
```
* See these metrics (they have new labels)
```
requests_resolved{operation_type="query",path="Query.hello",success="true",authenticated="false"} 4
requests_resolved_total{operation_type="query",path="Query.hello",success="true",authenticated="false"} 4
```

BTW, I copied the metrics result of the setup I mentioned earlier here: https://gist.github.com/aliok/ccea8e59f23c18ba1fc5966625fbab32

### DIFF
--- a/packages/voyager-metrics/src/metrics.ts
+++ b/packages/voyager-metrics/src/metrics.ts
@@ -15,13 +15,13 @@ const resolverTimingMetric = new Prometheus.Histogram({
 const resolverRequestsMetric = new Prometheus.Counter({
   name: 'requests_resolved',
   help: 'Number of requests resolved by server',
-  labelNames: ['operation_type', 'path']
+  labelNames: ['operation_type', 'path', 'success', 'authenticated']
 })
 
 const resolverRequestsTotalMetric = new Prometheus.Counter({
   name: 'requests_resolved_total',
   help: 'Number of requests resolved by server in total',
-  labelNames: ['operation_type', 'path']
+  labelNames: ['operation_type', 'path', 'success', 'authenticated']
 })
 
 const serverResponseMetric = new Prometheus.Histogram({
@@ -36,6 +36,19 @@ const conflictsMetric = new Prometheus.Counter({
   labelNames: ['operation_type', 'name']
 })
 
+const uniqueClientsMetric = new Prometheus.Counter({
+  name: 'unique_clients',
+  help: 'Number of unique clients'
+})
+
+const uniqueUsersMetric = new Prometheus.Counter({
+  name: 'unique_users',
+  help: 'Number of unique users'
+})
+
+const uniqueClientIds = new Set()
+const uniqueUserIds = new Set()
+
 /**
  *
  * @param app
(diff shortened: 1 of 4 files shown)


packages/voyager-server/src/voyagerResolver.ts
```diff
@@ -29,13 +29,11 @@ function voyagerResolverPartial (config: CompleteVoyagerConfig): ResolverWrapper
...
-          metrics.updateResolverMetrics(info, timeTook)
+          metrics.updateResolverMetrics(true, context, info, timeTook)
           auditLogger.logResolverCompletion('', true, obj, args, context, info)
         } catch (error) {
-          // we only publish time in success. const timeTook
-          // NOPE: const timeTook = Date.now() - resolverStartTime
-          // NOPE: updateResolverMetrics(info, timeTook)
-          // but, we audit log in case of failure too
+          const timeTook = Date.now() - resolverStartTime
+          metrics.updateResolverMetrics(false, context, info, timeTook)
```
aliok: I wanted to change the behavior here. Previously, when resolver failed we didn't record that and Prometheus didn't receive that. Now, after introduction of `success` field, we can do this. Prom will now get the info but it will also get the success/fail.

packages/voyager-metrics/src/metrics.ts
```diff
@@ -36,6 +36,19 @@ const conflictsMetric = new Prometheus.Counter({
...
+  name: 'unique_clients',
+  help: 'Number of unique clients'
+})
+
+const uniqueUsersMetric = new Prometheus.Counter({
+  name: 'unique_users',
+  help: 'Number of unique users'
+})
+
+const uniqueClientIds = new Set()
```
darahayes: I have a question @aliok. What is the plan to support multiple instances of this service? Is there some kind of aggregation in Prometheus/Grafana that will add these numbers together across instances?

I have a feeling that we will have false readings for `uniqueClientIds` and `uniqueUserIds` if we're adding these numbers together across multiple instances of services. Would it make more sense to log client/user IDs via audit logs or something else and then have that information counted and deduplicated at a top level by the metrics service?

packages/voyager-metrics/src/metrics.ts
```diff
@@ -72,16 +92,31 @@ export function updateResolverMetrics (resolverInfo: any, responseTime: number)
   resolverRequestsMetric
     .labels(
       resolverMappingType,
-      path
+      path,
+      success + '',
```
darahayes: Question: Is this hack to convert the value to a String?

aliok: Yes otherwise it won't compile

packages/voyager-metrics/src/metrics.ts
aliok: Great point. Prometheus service visits every pod and asks this information but still  your point is valid. Let me think about this.

packages/voyager-metrics/src/metrics.ts
wei-lee: I saw this on SO: https://stackoverflow.com/questions/51882134/prometheus-query-to-count-unique-label-values. So I think if we just sent a counter metric data with the label: `clientId=<value>`, and the we can just count it like this:

```
count( count by (clientId) (unique_clients))
```

packages/voyager-metrics/src/metrics.ts
wei-lee: I saw this on SO: https://stackoverflow.com/questions/51882134/prometheus-query-to-count-unique-label-values. So I think if we just sent a counter metric data with the label: `clientId=<value>`, and the we can just count it like this:

```
count( count by (clientId) (unique_clients))
```

aliok: Yes, this works. Changing the code and sending a new PR now.
thx @wei-lee

PR #110, opened 2019-02-15
fix: publish all examples to dockerhub

psturc:
@wtrocki @darahayes wdyt?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -80,7 +80,7 @@ jobs:
           docker_layer_caching: true
       - run: |
           cd examples
-          TAG=$CIRCLE_TAG ../scripts/publishExamples.sh basic
+          TAG=$CIRCLE_TAG ../scripts/publishExamples.sh auditLogging basic conflicts keycloak metrics restapi
 workflows:
   version: 2
   build_and_release:


wtrocki:
https://groups.google.com/forum/m/#!topic/aerogear/wiz53_l5qXo

wojta:
@psturc The script allows to do that, but I published only `basic` so far because of that it makes additional considerations for maintaining other ones.

psturc:
@wtrocki I needed some server docker image with `/metrics` endpoint and latest `@aerogear/metrics` dependency I could use for testing of sync-app-apb with metrics-apb and by looking at the aerogear dockerhub org, I noticed there are already  [3 "voyager-server-example-*"](https://hub.docker.com/u/aerogear) docker images present, but not the `*-metrics` one.
So I thought it would be good to have all examples from this repo released. I missed the mail thread you've mentioned, sorry.

I realised now that [this image](https://hub.docker.com/r/aerogear/voyager-server-example-task) should actually have `/metrics` endpoint, but it's pushed 2 weeks ago and not using latest `@aerogear/metrics` package. Are we planning some release automation for [ionic-showcase server](https://github.com/aerogear/ionic-showcase/tree/master/server)?

wtrocki:
@psturc We are planning to do continuous releases on the showcase app.

PR #129, opened 2019-02-28
Add reject method to conflict resolution

wtrocki:
## Motivation

Users should be able to reject client changes without writing data to the server.
With strategy serverWins we writing the same data back to the database. 
This is not needed and it can actually override the writes that happened in the time between reading and write. This method is just a helper to document this case and allow to reject changes made by the client explicitly. 

This approach will work with the client out of the box as for server-side resolution users will receive notification and can act on it.
More info in AEROGEAR-8614 

## Verification

Unit tests providing verification for this case. 
I have also tested my manually changing strategy in examples.
Reject method is a corner case so it is best to not include that in conflict example and just document this. 

@darahayes @jhellar Minor updates in docs will be needed after that change:
- Remove any occurrences of serverWins strategy
- Describe the reject method and purpose of this method

### DIFF
--- a/packages/voyager-conflicts/src/api/ObjectState.ts
+++ b/packages/voyager-conflicts/src/api/ObjectState.ts
@@ -41,6 +41,15 @@ export interface ObjectState {
    * @param serverState the current state of the object on the server
    * @param clientState the state of the object the client wishes to perform some mutation with
    */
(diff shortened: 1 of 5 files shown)


wtrocki:
Ready for review!

PR #164, opened 2019-05-10
Voyager Subscriptions Package

darahayes:
### Description

Related PRs:

* https://github.com/aerogear/aerogear-js-sdk/pull/346
* https://github.com/aerogear/ionic-showcase/pull/175

### To Do

- [x] Comments everywhere
- [x] Unit tests for the context combining functions in `voyager-subscriptions`
- [x] Unit tests for the onSubscriptionConnect function in `voyager-keycloak`
- [ ] An example of keycloak + subscriptions? Or just add subscriptions to the keycloak example? (cc @wtrocki @wei-lee @StephenCoady what are your thoughts on that?)
- [x] Update example smoke tests to do some trivial testing of the subscriptions example

##### Checklist

- [x] `npm test` passes
- [x] tests are included


packages/voyager-keycloak/src/KeycloakSecurityService.ts
```diff
@@ -90,9 +90,28 @@ export class KeycloakSecurityService implements SecurityService {
...
+    try {
+      await this.keycloak.grantManager.validateToken(token, 'Bearer')
+      return token
+    } catch (e) {
+      this.log.error(`Error validating token from connectionParam ${header}\n${e}`)
+      throw new Error(`Access Denied - ${e}`)
+    }
+  }
+
+  private getBearerTokenFromHeader(header: any) {
```
wtrocki: Original code pointed where we got this method. Let's put it here as well.

darahayes: This isn't taken from anywhere. I wrote it 😄 I will be putting some comments all over the place soon though.

wtrocki: I have seen that before. Oh wait...
https://github.com/keycloak/keycloak-nodejs-connect/blob/214d576f3b013977eee664ff06ba955e358b043d/stores/bearer-store.js#L20-L31

darahayes:
@wtrocki @StephenCoady @wei-lee So a summary of what this PR does:

* adds a `onSubscriptionConnect` function to the securityService interface. This function is implemented by a security service (such as keycloak) and it handles incoming subscription connections. Whatever is returned from this function is made available in the `context` inside the 
* implement onSubscriptionConnect in the KeycloakSecurityService. It essentially parses the token sent by the client and builds it into a `Token` object that has methods like `hasRole` etc. This can then be used by developers in their subscription resolvers to add RBAC to subscriptions.
* implement the `voyager-subscriptions` package. It does very little except provides an abstraction that allows different framework components (such as the security service) to add additional stuff to the context, while users can still provide their own `onConnect` function. The idea here is that we are as backwards compatible as possible with the regular SubscriptionServer API.
* unit tests for the keycloak `onSubscriptionConnect`
* unit tests for the context provider abstraction
* trivial example server showing the subscriptions package usage
* smoke test for that example.

examples/subscriptions/server.js
```diff
@@ -0,0 +1,52 @@
...
+    hello: String
+  }
+`
+
+// Resolver functions. This is our business logic
+const resolvers = {
+  Query: {
+    hello: (obj, args, context, info) => {
+      const message = args.name ? `Hello ${args.name}` : 'Hello world'
+      pubsub.publish(TOPIC, { hello: message })
```
wtrocki: @StephenCoady Do you think we can so some layer on top of that as separate ticket or is not worth the hassle now? Any simplifications on top of plain package that we can do as part of this epic?

packages/voyager-keycloak/src/KeycloakSecurityService.ts
```diff
@@ -65,9 +69,11 @@ export class KeycloakSecurityService implements SecurityService {
       store: memoryStore
     }))
 
-    this.keycloak = new Keycloak({
-      store: memoryStore
-    }, this.keycloakConfig)
+    if (!this.keycloak) {
+      this.keycloak = new Keycloak({
+        store: memoryStore
```
wtrocki: @darahayes Unrelated to this it using in-memory lib for session may be tech debt? 
Do you thin is worth issue?

packages/voyager-keycloak/src/KeycloakSecurityService.ts
```diff
@@ -90,9 +96,29 @@ export class KeycloakSecurityService implements SecurityService {
     }
   }
 
-  public async validateToken(token: string): Promise<boolean> {
-    const tokenObject = getTokenObject(token)
-    const result = await this.keycloak.grantManager.validateToken(tokenObject, 'Bearer')
-    return (result === tokenObject) ? true : false
+  public async onSubscriptionConnect(connectionParams: any, webSocket: any, context: any): Promise<any> {
+    const header = connectionParams.Authorization
+                  || connectionParams.authorization
```
wtrocki: backward compatibility? I think we should send token here and detect if object is string

packages/voyager-subscriptions/src/VoyagerSubscriptionContextProvider.ts
```diff
@@ -0,0 +1,97 @@
+// import { AuditLogger } from '@aerogear/voyager-audit'
+import { SecurityService } from '@aerogear/voyager-keycloak'
+
+import { VoyagerSubscriptionServerOptions } from './api'
+
+type OnConnectFunction = (connectionParams: any, webSocket: any, context: any) => any
```
wtrocki: Forgot to export?

packages/voyager-keycloak/src/KeycloakSecurityService.ts
StephenCoady: I was going to say if we interface connectionParams then we can type it properly and future proof it for whatever else we might need to add.

examples/subscriptions/server.js
StephenCoady: I'm not sure what you mean? Do you mean like we had in the showcase? Not sure we can simplify this for the example if thats what you mean

packages/voyager-subscriptions/src/VoyagerSubscriptionContextProvider.ts
```diff
@@ -0,0 +1,97 @@
...
+ * 'onConnect' function that allows developers to supply their own onConnect
+ * the way they would with plain Apollo. However this onConnect will be combined
+ * with our own onConnect which may do some things such as authentication/authorization
+ * The results from our onConnect and the developer's onConnect functions are combined together
+ * and are made available in the GraphQL context inside subscription resolvers
+ *
+ */
+export class VoyagerSubscriptionContextProvider {
+
+  public onConnectFunction: any
```
wtrocki: Missing type

wtrocki:
Tested only with the example app provided.

examples/subscriptions/server.js
wtrocki: Initial idea of the trello epic (if we look into it was to create subscription package that will be helping users to grasp the complexity of the thing. 
I'm curious if we actually can do more than this package now. (possibly not much apart from docs)

packages/voyager-keycloak/src/KeycloakSecurityService.ts
darahayes: Absolutely worth it

packages/voyager-subscriptions/src/VoyagerSubscriptionContextProvider.ts
darahayes: Fixed that now. Moved it into API.

packages/voyager-subscriptions/src/VoyagerSubscriptionContextProvider.ts
darahayes: I have changed this to `Function` because that's the type the `SubscriptionServer` uses.

packages/voyager-keycloak/src/KeycloakSecurityService.ts
darahayes: The way we parse this token is compatible with GraphQL playground and other similar technologies where there is usually a box to pass in some extra http headers. The same values in there are also passed in the subscription payload.

With regards to connectionParams, I'm not sure about trying to interface that as it literally can be anything. The client can pass objects, raw values like Strings and numbers, they can pass nothing at all. We can define an interface for what the connectionParams should look like when using auth, and then cast to it. That will allow us to have a nice development experience but you still need to do some checks at runtime.

wtrocki: Sorry for that. I totally forgot that we wanted to really keep GraphQL playground compatibility.

darahayes:
I have published a new dev release of the packages `0.6.1-dev-1` that include the changes from this PR

PR #167, opened 2019-05-17
Fix voyager-subscriptions version in examples

b1zzu:
### Description

Because version `0.5.1` doesn't exist for `@aerogear/voyager-subscriptions` we should use the version `0.5.2-dev.1` until the next release.

##### Checklist

- [ ] `npm test` passes
- [ ] tests are included

### DIFF
--- a/examples/package.json
+++ b/examples/package.json
@@ -24,7 +24,7 @@
     "@aerogear/voyager-metrics": "^0.5.1",
     "@aerogear/voyager-server": "^0.5.1",
     "@aerogear/voyager-tools": "^0.5.1",
-    "@aerogear/voyager-subscriptions": "^0.5.1",
+    "@aerogear/voyager-subscriptions": "^0.5.2-dev.1",
     "graphql-subscriptions": "1.1.0",
     "axios": "^0.18.0",
     "express": "^4.16.4",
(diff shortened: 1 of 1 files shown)


wtrocki:
We recommend to use lerna bootstrap or simply npm run bootstrap to get this working on local machine. Examples are designed to work with the unrelased versions from master but maybe this stopped working now

darahayes:
@b1zzu can you try running `npm run clean` from the top level directory and then run `npm run bootstrap.` Will the example server work?

I have this example merged but the package isn't released yet but it will be in the next day or so.

b1zzu:
Using `npm run bootstrap` works. But I was testing to deploy of the example app in OpenShift using the data sync template [1] , and because the template is using S2I in order to build the example it doesn't work in that context. (S2I use npm install)

1. https://github.com/aerogear/datasync-deployment/blob/master/openshift/datasync-http.yml

darahayes:
@b1zzu thanks for that info. I'm happy to merge this PR for the time being if it helps unblock any work you are doing on the openshift side. 👍

darahayes:
@b1zzu could you update it to the latest version? It's `0.6.1-dev-1`. That will be the proper version number once it's released. The 0.5.1 was an accidental publish 😄

b1zzu:
@darahayes I've updated it

PR #179, opened 2019-06-25
fix: ensure onSubscriptionConnect throws when no connectionParams supplied

darahayes:
I found a case where if keycloak auth is enabled on the server side but the client sends absolutely nothing in its connection params, an error is thrown like `cannot read property Authorization of undefined.` This is because the `onSubscriptionConnect` function trying to access some properties on the `connectionParams` when it is undefined.

The desired outcome of the client not being able to connect is achieved but the error message is wrong.

This PR checks for that case and also adds some extra tests for it.

##### Checklist

- [x] `npm test` passes
- [x] tests are included

### DIFF
--- a/packages/voyager-keycloak/src/KeycloakSecurityService.ts
+++ b/packages/voyager-keycloak/src/KeycloakSecurityService.ts
@@ -97,6 +97,9 @@ export class KeycloakSecurityService implements SecurityService {
   }
 
   public async onSubscriptionConnect(connectionParams: any, webSocket: any, context: any): Promise<any> {
+    if (!connectionParams || typeof connectionParams !== 'object') {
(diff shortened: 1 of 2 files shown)


packages/voyager-keycloak/src/KeycloakSecurityService.ts
```diff
@@ -97,6 +97,9 @@ export class KeycloakSecurityService implements SecurityService {
   }
 
   public async onSubscriptionConnect(connectionParams: any, webSocket: any, context: any): Promise<any> {
+    if (!connectionParams || typeof connectionParams !== 'object') {
```
wtrocki: Second check covers first as well.
```suggestion
    if (typeof connectionParams !== 'object') {
```

PR #192, opened 2019-07-12
fix: use published keycloak implementation

darahayes:
### Description

* Removes the majority of the business logic in the `voyager-keycloak` library.
* `voyager-keycloak` mostly just remaps things from `keycloak-connect-graphql` now
* There are a couple of hacks to ensure that the keycloak stuff will be included in both `context.auth` and `context.kauth`. The reason is that `voyager-server` wants `context.auth` to exist (idea is that we are not specific to keycloak on a top level) and `keycloak-connect-graphql` needs `context.kauth` to work properly so the hack there is to assingn the context Keycloak context objects on both `context.auth` and `context.kauth`

## TODO

@wtrocki we discovered that this `applyAuthMiddleware` function is really unsuitable for production now. I think we need to have a chat about that and figure out how to fix it properly. That might require lots of changes to docs and examples.

##### Checklist

- [ ] `npm test` passes
- [ ] tests are included


wtrocki:
Yep. I'm really glad you did not come back and changed upstream package from kauth to auth. It must be like that. The idea of using a separate security page was always strange here because most of the security packages like account.js can be used already with GraphQL and have directives and stuff. So we, in fact, introducing more work than simplifying things by this interface. The fact that we needed to change this interface 5 times (I have checked history) proving that implementing any other change that will conform to this interface can be tricky. 

IMHO this is good fix. Long term we can drop the entire keycloak package here.

wtrocki:
Please do not merge. Approved it but still need to verify this change now/

PR #216, opened 2019-12-10
Deprecate metrics

wtrocki:
Required to kickstart docs engagement:

- rename docs folder to docs
- move dev related docs to resources
- Remove metrics and audit logs (docs/examples - packages to follow up)

## What I would love to do:
Publish updated versions of the audit log packages. 
I would also remove conflict package as it makes no sense - WDYT?

# Motivation

 We going to need to invest more time into voyager-server - my recommendation will be to rename repo name to to datasync as soon as possible as we had cleared that name but we can still keep the voyager server packages. We will need to build this production ready example template that uses offix/graphback and has all the features done using offix and graphback.

# Suggested approach for the docs

Currently documentation got outdated. We have docs that are being used now by red hat consultants - graphback.dev and offix.dev pages. They are very specific to graphql elements.

In this repo we envision merging both packages into single app that can be pushed to OpenShift in zero time. This will invalidate a lot of documentation:

- Having template will mean that our way to deploy to openshift will change
- Many guides will be simplified.

## What to do for the moment

To avoid blocking work and keeping things outdated we should strip down integrately docs and link to offix.dev and graphback.dev docs same as we do for the keycloak etc.
I think we need to keep things lightweight and let team to build on that when we get the example app going. In this PR I have made some initial cleanup, but we can go even further.

Plan for the moment. Clear integrately docs from all graphql related stuff. 
Integrately developer cares only about the how to run things and not how graphql works. 
Have datasync docusaurus website that integrately can point on - this will merge all projects together into single runtime that gives ultimate experience. 

Team will build docs now more frequently and maintain it better as it is directly inside repository.

## Main questions

- Do we need to keep antora format? I think we can make it work for both docusaurus and antora 
- Is simplifiing integrately docs possible - it reduces amount of work and time we will require to get it. Team can execute all of that without bothering docs team.
- Can we link to graphback/offix docs?
- Can we have upstream website here or it is not worth the hassle?


wtrocki:
@pwright See comment above with the plan. I would like to keep this PR as a way to clean up the repository. 

Above I have posted some context, but long story short:

- Can we keep Antora going but clear up some docs to reflect stuff we support that is related to integrately premise (simplification)
- Can we point people to upstream docs - offix.dev and graphback.dev - our plan would be to eventually move all of the complexity of using these packages into this repository but.. for the moment we need quick wins and get that published

wtrocki:
Also plan for showcase should be clear: 
https://github.com/aerogear/ionic-showcase/pull/283

wtrocki:
@pwright @darahayes Would it be possible to get approval on this?

PR #219, opened 2019-12-12
Docs refined

wtrocki:


docs/integreatly/modules/ROOT/pages/ds-realtime.adoc
```diff
@@ -3,7 +3,7 @@ include::_partials/attributes.adoc[]
 // end::excludeDownstream[]
 
 [id='realtime_{context}']
-= Supporting real-time updates in your mobile app
+= Subscriptions - Supporting real-time updates in your mobile app
```
wtrocki: I know that this is controversial but this is good example where docs obfuscate real knowledge. 
Subscriptions as single word means `Supporting real-time updates in your mobile app` so instead of long title we can have just subscriptions - same as in other chapters

docs/integreatly/modules/ROOT/pages/_partials/data-sync/openshift-templates.adoc
```diff
@@ -21,12 +21,7 @@ Server side requires client application available on github link:https://github.
 
 Note: Use the Showcase server template for demo purposes only. Do not use this template in a production environment.
 
+When running on Red Hat Managed Integration templates will be available by Searching `DataSync`  in OpenShift catalog
 
-// tag::excludeUpstream[]
-When running on Red Hat Managed Integration templates will be available in *Mobile* > *App*  category in OpenShift catalog
-// end::excludeUpstream[]
+Source code for templates along with documentation can be found in link:https://github.com/aerogear/datasync-deployment[datasync-deployment] repository.
```
wtrocki: This is where we tell people - hey you can find this templates already on integrately,
but if you want to change app template (and people will) there is source code

PR #228, opened 2019-12-16
add showcase docs

pwright:
reduced docs from mobile-docs


docs/showcase/modules/ROOT/pages/push-notifications.adoc
```diff
@@ -0,0 +1,8 @@
+
+
+= {unifiedpush-service}
+
+
+. Send a push notification as described in
+link:{push-guide-link}#sending[Sending a Push Notification]
```
wtrocki: What we do with this? I think that push should link to showcase same with sync. Not the other way around

wtrocki:
Generally after reviewing the content I think that amount of the files might not justify having this a separate thing. Maybe we can have it merged into or get into some consenus how we link this with push and sync together

wtrocki:
Removed setup docs:
None of this info is valid. Showcase is available as part of the template that is described in sync docs (and might be in push) and content there is not reflecting any setup - all was focued on MDC and currently showcase is automatically configured so there is no extra work needed to get it running unless we do Auth. For auth we have dedidacted walkthough

docs/showcase/modules/ROOT/pages/_partials/attributes.adoc
```diff
@@ -0,0 +1,76 @@
...
+:service-name:
+
+:mobile-client: Mobile App in MDC
+:multi-mobile-client: Mobile Apps in MDC
+:short-mobile-client: Mobile App
+:mobile-client-openshift: Mobile Client in your OpenShift project
+:mobile-cli: Mobile CLI
+
+// Metrics Service
+:metrics-service: Mobile Metrics
```
wtrocki: MOst of those are no longer used and valid

docs/showcase/modules/ROOT/pages/_partials/attributes.adoc
```diff
@@ -0,0 +1,76 @@
...
+:unifiedpush-service: Push Notifications
+:push-ui: Unified Push Admin UI
+:push-notification: push notification
+
+// Build Service
+:mobile-ci-cd-service: Mobile CI/CD
+:mobilecicd-ui: Jenkins UI
+
+// Device Security
+:device-security-service: Device Checks
```
wtrocki: Another example

docs/showcase/modules/ROOT/pages/index.adoc
```diff
@@ -0,0 +1,26 @@
...
+include::_partials/attributes.adoc[]
+
+:toc:
+
+= Using the Showcase App
+
+include::service-setup.adoc[leveloffset=1,tags={include-tags}]
+
+
+// tag::excludeDownstream[]
```
wtrocki: WOW. so if that is excluded downstream let's remove it then. 
Information there is kinda common problem and should be documented in stackoverflow rather than in our docs.

docs/showcase/modules/ROOT/pages/index.adoc
```diff
@@ -0,0 +1,26 @@
...
+[#certs]
+
+include::proc_self-signed-certs.adoc[leveloffset=1,tags={include-tags}]
+// end::excludeDownstream[]
+
+[#features]
+== Demonstrating the ShowCase App features
+
+include::identity-management-demo.adoc[leveloffset=2,tags={include-tags}]
```
wtrocki: Let's keep in mind that we no longer have indentity-management service so we need to keep their official name - keycloak/SSO

PR #245, opened 2020-01-13
remove duplicate image

pwright:
very difficult to test, because problem only occurs from user-docs-site repo rendering on netlify


pwright:
@joaedwar fixed the rendering by modifying user-docs-site, but this PR is still necessary to remove duplicate image

